Issue #36, opened 2016-09-16
Base executor requirements and execution contexts


chriskohlhoff:
The base executor requirements should include:
- Copyable, Destructible
- EqualityComparable
- Noexcept guarantees (no copy constructor, destructor, equality operator, swap operation shall exit via an exception)
- `context_type` typedef and `context_type& context()` member function.

For example:

```
class thread_pool
{
public:
  class executor_type;
  // ...
};

class thread_pool::executor
{
public:
  executor(const executor&) noexcept;
  typedef thread_pool context_type;
  context_type& context() const;
  friend bool operator==(const executor& a, const executor& b) noexcept;
  friend bool operator!=(const executor& a, const executor& b) noexcept;
  // ...
};
```

At the base level, almost no requirements are placed on the type represented by `context_type`. In a generic context the only thing you can do with it is compare the address. However, refinements of the executor requirements can place additional requirements on `context_type`.

Example - using thread pool executors that may refer to different thread pools:

```
template <class Ex1, class Fn1, class Ex2, class Fn2>
void my_algorithm(Ex1 ex1, Fn1 fn1, Ex2 ex2, Fn2 fn2)
{
  if (std::addressof(ex1.context()) == std::addressof(ex2.context()))
  {
    // Some optimisation based on the knowledge that both executors
    // utilise the same underlying execution context.
  }
  else
  {
    // ...
  }
}

int main()
{
  thread_pool pool1;
  thread_pool pool2;

  my_algorithm(pool1.get_executor(), []{}, pool2.get_executor(), []{});
  my_algorithm(pool1.get_executor(), []{}, pool1.get_executor(), []{});

  pool1.join();
  pool2.join();
}
```

Example - where executors are adapted, but still refer to the same underlying type of execution context:

```
template <class Ex1, class Fn1, class Ex2, class Fn2>
void my_algorithm(Ex1 ex1, Fn1 fn1, Ex2 ex2, Fn2 fn2)
{
  if (std::addressof(ex1.context()) == std::addressof(ex2.context()))
  {
    // Some optimisation based on the knowledge that both executors
    // utilise the same underlying execution context.
  }
  else
  {
    // ...
  }
}

template <class Ex>
class logging_executor
{
public:

  logging_executor(Ex ex) noexcept : ex_(ex) {}

  typedef typename Ex::context_type context_type;
  execution_context& context() const noexcept { return ex_.context(); }

  template <class F, class Alloc>
  void post(F f, const Alloc& a)
  {
[...]

PR #40, opened 2016-09-29
Add ExecutionContext and BaseExecutor requirements.

chriskohlhoff:
To address #36.


jaredhoberock:
Thanks Chris. It looks like `OneWayExecutor` lost the `EqualityComparable` requirement. Was that intentional? I was expecting `BaseExecutor` to receive the `EqualityComparable` requirement, but it looks like only `ExecutionContext` got it.

chriskohlhoff:
It should have been covered by the ==/!= rows in BaseExecutor's requirements table, but I've just added it to BaseExecutor's list of requirements to be sure.

jaredhoberock:
Thanks Chris, makes sense to me.

PR #64, opened 2016-10-14
Add an introduction to the paper

jaredhoberock:
Eliminate the old TODO.md


jaredhoberock:
Custom executor idea: adaptation of another executor that would add logging
OpenMP executor example: Start with one-way executor (OpenMP task), then show how to add an additional customization point for bulk-two-way (omp parallel for).

PR #95, opened 2016-12-19
Issue 94

jaredhoberock:
This pull request introduces `executor_execution_mapping_category`, `executor_blocking_category`, and related wording.

### DIFF
--- a/README.md
+++ b/README.md
@@ -334,12 +334,12 @@ namespace execution {
 
   struct blocking_execution_tag {};
   struct possibly_blocking_execution_tag {};
-  struct nonblocking_execution_tag {};
+  struct non_blocking_execution_tag {};
 
-  template<class Executor> struct executor_blocking_category;
+  template<class Executor> struct executor_execute_blocking_category;
 
   template<class Executor>
-    using executor_blocking_category_t = typename executor_blocking_category<Executor>::type;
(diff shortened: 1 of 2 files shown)


jaredhoberock:
@chriskohlhoff - If you have time, I'd appreciate feedback on this pull request.

chriskohlhoff:
As discussed on saturday, can we call it executor_execute_blocking_category (to make it clearer that it applies only to the `*execute()` functions?

I think the phrasing of "`other_execution_mapping_tag` indicates that execution agents created by a component may not be mapped onto threads of execution." isn't exactly right -- "may not be" sounds like it isn't allowed to use normal threads of execution at all. Presumably we want to use this classification for executors that could use, say, a GPU *or* a thread. Something like "... may be mapped on to ??? other than a thread of execution". Not sure what to put for the ??? though.

I'm not totally comfortable with the use of "thread of execution" phrase (may just take some getting used to). For now can we include a comment that we mean this as something like as-if a std thread?

I think you accidentally a word in the phrase "completion of execution created" which occurs in several places, presumably "agent"?

Also minor naming suggestion `nonblocking_...` -> `non_blocking_...` (the latter form has precedent in the networking TS).

jaredhoberock:
Thanks, good suggestions.

PR #101, opened 2017-01-30
Generalized executors

chriskohlhoff:
This is not ready to be merged, but I'm opening this pull request to give us a place to discuss the changes. Also, some people on the call said they could help with wording for the other customization points, so this might be the place to do that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -490,6 +490,87 @@ Table: (Base executor requirements) \label{base_executor_requirements}
 | `x1 != x2` | `bool` | Same as `!(x1 == x2)`. |
 | `x1.context()` | `E&` or `const E&` where `E` is a type that satisfies the `ExecutionContext` requirements. | Shall not exit via an exception. The comparison operators and member functions defined in these requirements (TODO and the other executor requirements defined in this Technical Specification) shall not alter the reference returned by this function. |
 
+### Requirements on execution functions
+
+An execution function is a member function of the form:
+
+    x.e(...)
+
+or a non-member function of the form:
+
+    e(x, ...)
+
+where `x` denotes an executor object. The function name `e`, and the remaining arguments, are determined according to the properties and semantics of the execution function, as described below.
+
+#### Naming of execution functions
+
+The name of an execution function is determined by the combination of properties and semantics that apply to the function and to the execution agents created by it. A word or prefix is associated with each property, and these may be concatenated in the order below.
+
+| Cardinality | Directionality | Blocking semantics |
+|-------------|----------------|-------------------|
+| "" or "bulk_" | "" or "sync_" or "async_" | "execute" or "post" or "defer" |
+
+#### Cardinality
+
+The cardinality property of an execution function may be one of the following:
+
+* *Single:* The execution function permits submission of a single function object to the executor. The names of execution functions having single cardinality do not have an associated prefix.
+* *Bulk:* The execution function allows submission of multiple function objects in a single invocation, with the number determined at runtime. The names of execution functions having bulk cardinality have the prefix `bulk_`.
+
+##### Requirements on execution functions of single cardinality
+
+In the Table below, `x` denotes a (possibly const) executor object of type `X`, `e` denotes the name of the execution function, `f` denotes a function object of type `F&&` callable as `DECAY_COPY(std::forward<F>(f))()` and where `decay_t<F>` satisfies the `MoveConstructible` requirements, and `a` denotes a (possibly const) value of type `A` satisfying the `ProtoAllocator` requirements.
+
+| Form | Semantics |
+|------|-----------|
+| `x.e(f)` | Equivalent to `x.e(f, executor_default_allocator_v<X>(x))`. |
+| `e(x,f)` | Equivalent to `e(x, f, executor_default_allocator_v<X>(x))`. |
+| `x.e(f,a)` <br/>`e(x,f,a)` | Creates a weakly parallel execution agent which invokes `DECAY_COPY( std::forward<F>(f))()` at most once, with the call to `DECAY_COPY` being evaluated in the thread that called `execute`. <br/><br/>May block forward progress of the caller until `DECAY_COPY( std::forward<F>(f))()` finishes execution. <br/><br/>Executor implementations should use the supplied allocator (if any) to allocate any memory required to store the function object. Prior to invoking the function object, the executor shall deallocate any memory allocated. [*Note:* Executors defined in this Technical Specification always use the supplied allocator unless otherwise specified. *--end note*] | *Synchronization:* The invocation of `execute` synchronizes with (C++Std [intro.multithread]) the invocation of `f`.|
+
+##### Requirements on execution functions of bulk cardinality
+
+*TODO*
+
+#### Directionality
+
+The directionality property of an execution function may be one of the following:
+
+* *One-way:* The execution function creates execution agents without a channel for awaiting the completion of a submitted function object or for obtaining its result. *Note:* That is, the executor provides fire-and-forget semantics. *--end note*] The names of execution functions having one-way directionality do not have an associated prefix.
+* *Synchronous two-way:* The execution function blocks until execution of the submitted function is complete, and returns the result. The names of execution functions having synchronous two-way directionality have the prefix `sync_`.
+* *Asynchronous two-way:* The execution function provides a *TODO* future-like channel for awaiting the completion of a submitted function object and obtaining its result. The names of execution functions having asynchronous two-way directionality have the prefix `async_`.
+
+##### Requirements on execution functions of one-way directionality
+
+In the Table below, `x` denotes a (possibly const) executor object of type `X`, and `e` denotes the name of the execution function.
+
+| Form | Return Type |
+|------|-------------|
+| `x.e(...)` | `void` |
+| `e(x,...)` | `void` |
+
+##### Requirements on execution functions of synchronous two-way directionality
+
+*TODO*
+
+##### Requirements on execution functions of asynchronous two-way directionality
+
+*TODO*
+
+#### Blocking semantics
+
+The blocking semantics of an execution function may be one of the following:
+
+* *Potentially blocking:* The execution function may block the caller pending completion of the submitted function objects. Execution functions having potentially blocking semantics are named `execute`.
+* *Non-blocking:* The execution function shall not block the caller pending completion of the submitted function objects. Execution functions having non-blocking semantics are named `post` or `defer`.
+
+##### Requirements on execution functions having potentially blocking semantics
+
+*TODO*
+
+##### Requirements on execution functions having non-blocking semantics
+
+*TODO*
+
 ### `OneWayExecutor` requirements
 
 The `OneWayExecutor` requirements form the basis of the one-way executor concept taxonomy. This set of requirements specifies operations for creating execution agents without a channel for awaiting the completion of a submitted function object and obtaining its result. [*Note:* That is, the executor provides fire-and-forget semantics. *--end note*]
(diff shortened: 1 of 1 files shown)


jaredhoberock:
Thanks, Chris. It looks like the customization points are now "customization point objects" with the form from the Ranges TS.

It doesn't look like the change has been made yet, but I take it the idea is to replace the free functions beginning with

```
  // Executor customization points:

  template<class OneWayExecutor, class Function>
    void execute(const OneWayExecutor& exec, Function&& f);
```

in favor of the newer-style customization points appearing earlier in the synopsis? That's fine with me, but just want to be sure I understand.

jaredhoberock:
Here's a naming idea. The traits with names like `has_execute` seem to me like they're intended to indicate whether or not an executor type is compatible with a particular customization point. In this case, `has_execute<T>` indicates that the executor type `T` can be used with `execution::execute()`. Perhaps these are `can_` traits rather than `has_` traits?

I think that tweak would make the meaning of the categories a bit clearer. For example, a `OneWayExecutor` is any executor which can be used with `execution::execute()`, even if `has_execute_member_v` is false.

chriskohlhoff:
> Thanks, Chris. It looks like the customization points are now "customization point objects" with the form from the Ranges TS.

Sorry, yes, for some reason I had it stuck in my mind as an intended change after the last call... not sure how I got that idea. Happy to continue converting them to this form assuming no showstoppers. Which do you prefer?

> It doesn't look like the change has been made yet, but I take it the idea is to replace the free functions beginning with
> 
>     // Executor customization points:
> 
>     template<class OneWayExecutor, class Function>
>       void execute(const OneWayExecutor& exec, Function&& f);
> 
> in favor of the newer-style customization points appearing earlier in the synopsis? That's fine with me, but just want to be sure I understand.

Correct. I've left them for now to make it easier to convert each function's semantics across to the new form.

> Here's a naming idea. The traits with names like has_execute seem to me like they're intended to indicate whether or not an executor type is compatible with a particular customization point. In this case, has_execute<T> indicates that the executor type T can be used with execution::execute(). Perhaps these are can_ traits rather than has_ traits?
> 
> I think that tweak would make the meaning of the categories a bit clearer. For example, a OneWayExecutor is any executor which can be used with execution::execute(), even if has_execute_member_v is false.

The `can_` form sounds good to me.

jaredhoberock:
> Sorry, yes, for some reason I had it stuck in my mind as an intended change after the last call... not sure how I got that idea. Happy to continue converting them to this form assuming no showstoppers. Which do you prefer?

I'm ambivalent about what form the customization points take. No strong preference right now. The advantage of using the new style is that it avoids all the issues people have identified over the years with the old-style customization points like `swap()`. We can also repurpose all the work done in the Ranges TS to generate the specifications. The disadvantage of the new-style form is that it can be difficult to understand the semantics of the customization points at a glance when they are not simple function templates.

Do you have an idea of what the specification of something like `execution::execute()` will look like as a customization point object?

chriskohlhoff:
> Do you have an idea of what the specification of something like execution::execute() will look like as a customization point object?

That's the one (and only one so far) that I have converted. Search for the "Customization points" heading in the wording.

jaredhoberock:
Here's the proposed new wording for `execution::execute()`, for reference:

> ```
> execute
> 
> namespace {
>   constexpr unspecified execute = unspecified;
> }
> ```
> The name `execute` denotes a customization point. The effect of the expression `std::experimental::concurrency_v2::execution::execute(E, F)` for some expressions `E` and `F` is equivalent to:
> 
> `(E).execute(F)` if `has_execute_member_v<decay_t<decltype(E)>>` is true.
> 
> Otherwise, `execute(E, F)` if that expression satisfies the syntactic requirements for a one-way, potentially blocking execution function of single cardinality, with overload resolution performed in a context that includes the declaration void `get_trivial_executor(auto&) = delete;` and does not include a declaration of `std::experimental::concurrency_v2::execution::execute`.
> 
> Otherwise, `std::experimental::concurrency_v2::execution::execute(E, F)` is ill-formed.
> 
> Remarks: Whenever `std::experimental::concurrency_v2::execution::execute(E, F)` is a valid expression, that expression satisfies the syntactics requirements for a one-way, potentially blocking execution function of single cardinality.

Could you explain what's going on in the Otherwise, clause? After a first reading, it looks like it calls the `.execute()` member if possible, and if not, tries the `execute()` free function. What does the bit about `get_trivial_executor()` do?

I was expecting the executor to be adapted if neither the member `.execute()` nor the free function exists, but it doesn't appear that that's what happens.

chriskohlhoff:
On Thu, Feb 2, 2017, at 07:48 AM, Jared Hoberock wrote:

> Could you explain what's going on in the Otherwise, clause?
> After a first
> reading, it looks like it calls the `.execute()` member if
> possible, and
> if not, tries the `execute()` free function.



Yes if both those fail the expression s I'll formed. 



> What does the bit about

> `get_trivial_executor()` do?



That's some copy/paste I forgot to rename. It should be 'execute'.



> I was expecting the executor to be adapted if neither the member

> `.execute()` nor the free function exists, but it doesn't appear that
> that's what happens.



The execute function is the most fundamental so there's nothing to
adapt. This is equivalent to what we had for our existing definition of
execution::execute.


Perhaps async_execute would be a more appropriate example for
evaluating this? I'll push something for that soon. For async_execute
the steps would be:


* use equivalently named member if present

* otherwise, use equivalently named free function if present

* otherwise adapt using execution::execute if well formed ie
  can_execute is true
* otherwise ill formed

jaredhoberock:
Here's something like what I had been envisioning for wording the customization points. The idea would be to specify the adaptation which happens when neither the member function nor free functions exist. In most cases, the adaptation would simply apply other "nearby" customization point. I'll use `sync_execute` to demonstrate. Let me know if this is compatible with your thinking.

> ```
> sync_execute
> 
> namespace {
>   constexpr unspecified sync_execute = unspecified;
> }
> ```
> 
> The name `sync_execute` denotes a customization point. The effect of the expression `std::experimental::concurrency_v2::execution::execute(E,F)` for some expresssions `E` and `F` is equivalent to:
> 
> `(E).sync_execute(F)` if `has_sync_execute_member_v<decay_t<decltype(E)>>` is `true`.
> 
> Otherwise, `sync_execute(E,F)` if... (that expression using the free function form is OK)
> 
> Otherwise, `std::experimental::concurrency_v2::execution::async_execute(E,F).get()`.

In other words, in most cases there is a fallback adaptation which gets applied, which is always the application of some customization point. The particular customization point to prefer in the adaptation would be whichever passes along the most information. There might be more than one fallback case, and selecting which case to apply would depend on the traits of the executor type.

jaredhoberock:
> Perhaps async_execute would be a more appropriate example for
> evaluating this? I'll push something for that soon. For async_execute
> the steps would be:
> 
> 
> * use equivalently named member if present
> 
> * otherwise, use equivalently named free function if present
> 
> * otherwise adapt using execution::execute if well formed ie
>   can_execute is true
> * otherwise ill formed

Thanks, this makes sense to me.

I wonder if a free function should be preferred to a member function when both exist? Here's the reasoning:

Suppose I use a library which provides some executor:

```
namespace library
{

struct executor
{
  template<class Function>
  void execute(Function&& f);

  ...
};

}
```

Suppose I want to use `library::executor`, but I don't like the way `library::executor::execute()` works and wish to provide my own customization. I could try to do that by introducing the free function `library::execute(library::executor, f)` as a third-party. However, it seems like this customization would not be chosen by `execution::execute()` because the existing member function `.execute()` takes priority.

I'm not sure which is the better policy; just throwing it out there.

chriskohlhoff:
> I wonder if a free function should be preferred to a member function when both exist?

I'd have to think it over more, but intuitively it makes me nervous since you'd be injecting functionality into a namespace that doesn't really belong to you. Other parts of that library might already make use of that library via execution::execute() and expect it to behave its existing way. Maybe this is something better addressed by an adapter?

jaredhoberock:
Thanks for adding the specification for `sync_execute()`, it makes sense to me.

I wonder if a set of type traits for detecting the existence of free functions (e.g., `has_async_execute_free_function_v` et al.) would be helpful in collapsing clauses like this:

> * Otherwise, `sync_execute(E, F)` if that expression satisfies the syntactic requirements for a synchronous two-way, potentially blocking execution function of single cardinality, with overload resolution performed in a context that includes the declaration `void sync_execute(auto&, auto&) = delete;` and does not include a declaration of `std::experimental::concurrency_v2::execution::sync_execute`.

The result would be a set of detectors for member functions, free functions, and whether or not each customization point is available.

AerialMantis:
Thanks Chris, the member function and customization point tables looks great.

I've added the rest of the entries and written out the rest of the parameter requirements for the customization point table and i've added the execution functions which take an allocator parameter to the conditions.

I was also thinking that we may want these traits to have a requirement on the execution behaviours associated as well, though I wasn't sure as that's not something which can be enforced at compile-time.

-----

### Customization point type traits

    template<class T> struct can_execute;
    template<class T> struct can_post;
    template<class T> struct can_defer;
    template<class T> struct can_sync_execute;
    template<class T> struct can_async_execute;
    template<class T> struct can_async_post;
    template<class T> struct can_async_defer;
    template<class T> struct can_then_execute;
    template<class T> struct can_bulk_execute;
    template<class T> struct can_bulk_sync_execute;
    template<class T> struct can_bulk_async_execute;
    template<class T> struct can_bulk_then_execute;

This sub-clause contains templates that may be used to query the properties of a type at compile time. Each of these templates is a UnaryTypeTrait (C++Std [meta.rqmts]) with a BaseCharacteristic of `true_type` if the corresponding condition is true, otherwise `false_type`.

In the Table below,
* `t` denotes a (possibly const) executor object of type `T`,
* `e` denotes the name of the execution function,
* `f` denotes a function object of type `F&&` callable as `DECAY_COPY(std::forward<F>(f))()`, where `decay_t<F>` satisfies the `MoveConstructible` requirements.
* `bf` denotes a function object of type `F&&` callable as `DECAY_COPY(std::forward<F>(f))(i, s)`,
 * where `i` denotes an object whose type is `executor_index_t<X>`,
 * where `s` denotes an object whose type is `S` and
 * where `decay_t<F>` satisfies the `MoveConstructible` requirements,
* `rf` denotes a `CopyConstructible` function object with one argument whose result type is `R`,
* `sf` denotes a `CopyConstructible` function object with one argument whose result type is `S`,
* `pred` denotes a future object whose result is `pr` and
* `a` denotes a (possibly const) value of type `A` satisfying the `ProtoAllocator` requirements.

| Template                   | Conditions           | Preconditions  |
|----------------------------|---------------------|----------------|
| `template<class T>` <br/> `struct can_execute` | The expressions `std::experimental::concurrency_v2::execution::execute(t, f)` and `std::experimental::concurrency_v2::execution::execute(t, f, a)` are well-formed. | `T` is a complete type. |
| `template<class T>` <br/>`struct can_post` | The expressions `std::experimental::concurrency_v2::execution::post(t, f)` and `std::experimental::concurrency_v2::execution::post(t, f, a)` are well-formed. | `T` is a complete type. |
| `template<class T>` <br/>`struct can_defer` | The expressions `std::experimental::concurrency_v2::execution::defer(t, f)` and `std::experimental::concurrency_v2::execution::defer(t, f, a)` are well-formed. | `T` is a complete type. |
| `template<class T>` <br/>`struct can_sync_execute` | The expressions `std::experimental::concurrency_v2::execution::sync_execute(t, f)` and `std::experimental::concurrency_v2::execution::sync_execute(t, f, a)` are well-formed. | `T` is a complete type. |
| `template<class T>` <br/>`struct can_async_execute` | The expressions `std::experimental::concurrency_v2::execution::async_execute(t, f)` and `std::experimental::concurrency_v2::execution::async_execute(t, f, a)` are well-formed. | `T` is a complete type. |
| `template<class T>` <br/>`struct can_async_post` | The expressions `std::experimental::concurrency_v2::execution::async_post(t, f)` and `std::experimental::concurrency_v2::execution::async_post(t, f, a)` are well-formed. | `T` is a complete type. |
| `template<class T>` <br/>`struct can_async_defer` | The expressions `std::experimental::concurrency_v2::execution::async_defer(t, f)` and `std::experimental::concurrency_v2::execution::async_defer(t, f, a)` is well-formed. | `T` is a complete type. |
| `template<class T>` <br/>`struct can_then_execute` | The expressions `std::experimental::concurrency_v2::execution::then_execute(t, f, pred)` and `std::experimental::concurrency_v2::execution::then_execute(t, f, pred)` are well-formed. | `T` is a complete type. |
| `template<class T>` <br/>`struct can_bulk_execute` | The expression `std::experimental::concurrency_v2::execution::bulk_execute(t, bf, s, rf, sf)` is well-formed. | `T` is a complete type. |
| `template<class T>` <br/>`struct can_bulk_sync_execute` | The expression `std::experimental::concurrency_v2::execution::bulk_sync_execute(t, bf, s, rf, sf)` is well-formed. | `T` is a complete type. |
| `template<class T>` <br/>`struct can_bulk_async_execute` | The expression `std::experimental::concurrency_v2::execution::bulk_async_execute(t, bf, s, rf, sf)` is well-formed. | `T` is a complete type. |
| `template<class T>` <br/>`struct can_bulk_then_execute` | The expression `std::experimental::concurrency_v2::execution::bulk_then_execute(t, bf, s, pred, rf, sf)` is well-formed. | `T` is a complete type. |

-----

I've put together a table we could use at the end of the **requirements on execution functions** section, this describes the various combinations of executor properties and what the corresponding member function an customization points would be. I think this could be useful to highlight the combinations which are invalid such as a non-blocking sync_ execution function. This also raises the question of whether we want to have bulk non-blocking execution functions; `bulk_post`, `bulk_defer`, `bulk_async_post` and bulk_async_defer`. What does everyone think?

-----

#### Execution function Combinations

The table below describes the execution member functions and non-member functions that can be supported by an executor category via various combinations of the execution function requirements.

| Cardinality | Directionality | Blocking semantics | Member function | Customization point |
| ------------ | -------------- | --------------------- | ------------------- | ---------------------- |
| Single | One-way | Potentially blocking | `x.execute(f)` <br/> `x.execute(f, a)` | `execute(x, f)` <br/> `execute(x, f, a)` |
| Single | One-way | Non-blocking | `x.post(f)` <br/> `x.post(f, a)` <br/> `x.defer(f)`  <br/> `x.defer(f, a)` | `post(x, f)` <br/> `post(x, f, a)` <br/> `defer(x, f)`  <br/> `defer(x, f, a)` |
| Single | Two-way synchronous | Potentially blocking | `x.sync_execute(f)` <br/> `x.sync_execute(f, a)` | `sync_execute(x, f)` <br/> `sync_execute(x, f, a)` |
| Single | Two-way synchronous | Non-blocking | NA | NA |
| Single | Two-way asynchronous | Potentially blocking | `x.async_execute(f)` <br/> `x.async_execute(f, a)`  <br/> `x.then_execute(f, pred)` <br/> `x.then_execute(f, pred, a)` | `async_execute(x, f)` <br/> `async_execute(x, f, a)`  <br/> `then_execute(x, f, pred)` <br/> `then_execute(x, f, pred, a)` |
| Single | Two-way asynchronous | Non-blocking | `x.async_post(f)` <br/> `x.async_post(f, a)` | `async_post(x, f)` <br/> `x.async_post(x, f, a)` |
| Bulk | One-way | Potentially blocking | `x.bulk_execute(f, s, sf)` | `bulk_execute(x, f, s, sf)` |
| Bulk | One-way | Non-blocking | ? | ? |
| Bulk | Two-way synchronous | Potentially blocking | `x.bulk_sync_execute(f, s, rf, sf)` | `bulk_sync_execute(x, f, s, rf, sf)` |
| Bulk | Two-way synchronous | Non-blocking | ? | ? |
| Bulk | Two-way asynchronous | Potentially blocking | `x.bulk_async_execute(f, s, rf, sf)` <br/> `x.bulk_then_execute(f, s, pred, rf, sf)` | `bulk_async_execute(x, f, s, rf, sf)` <br/> `bulk_then_execute(x, f, s, pred, rf, sf)` |
| Bulk | Two-way asynchronous | Non-blocking | ? | ? |

-----

I also started working on the wording for the other sections of **requirements on execution functions** though i've not finished that yet, so far i've only started on the bulk cardinality.

chriskohlhoff:
> I've added the rest of the entries and written out the rest of the parameter requirements for the customization point table and i've added the execution functions which take an allocator parameter to the conditions.

Thanks Gordon, will incorporate these changes. You've just reminded me about the allocator parameter as well. Will have to consider how to incorporate it into the customization point specificiation as cleanly as possible.

jaredhoberock:
I think we're all happy with the direction of this pull request. I'll go ahead and merge it to make it easier to work on the changes in tandem.

PR #103, opened 2017-02-02
modifications to thread_pool wording to change to static_thread_pool

ccmysen:
Making a few wording changes to the thread pool to rename and update some of the comments/justification.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1709,7 +1709,6 @@ class static_thread_pool
     class executor_type;
     
     // construction/destruction
-    static_thread_pool();
     explicit static_thread_pool(std::size_t num_threads);
     
     // nocopy
@@ -1758,13 +1757,6 @@ not introduce data races as a result of concurrent calls to those functions
 from different threads of execution.
 
 #### Construction and destruction
-
-```
-static_thread_pool();
-```
-
-*Effects:* Constructs a `static_thread_pool` object with an implementation defined
-number of threads of execution, as if by creating objects of type `std::thread`.
 
 ```
 static_thread_pool(std::size_t num_threads);
--- a/README.md
+++ b/README.md
@@ -1696,7 +1696,7 @@ This class represents a statically sized thread pool as a common/basic resource
 type. This pool is capable of being grown by attaching threads to the pool but
 it will not change size in any automatic way. This pool provides an
 effectively unbounded input queue and as such calls to add tasks to a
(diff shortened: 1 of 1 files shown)


jaredhoberock:
Thanks Chris. This looks OK to me -- no functional changes, just a rename and a tweak to the explanatory text. I'll merge after @chriskohlhoff has a chance to weigh in.

chriskohlhoff:
I like the name `static_thread_pool`. It suggests the other behaviour should be called `dynamic_thread_pool`.

I wonder if we should delete the note:

> (Note: The implementation is encouraged, but not required, to use the attached thread to execute submitted function objects. ...)

as a static thread pool should use the threads provided by the user? It lets us say that a static thread pool will use exactly the concurrency provided by the user. A dynamic thread pool would support this "encouraged" behaviour instead.

I also wonder if we should remove the default constructor, and make specification of the number of threads explicit?

chriskohlhoff:
Also we can delete the question:

> do we want to allow 0?

as 0 threads at construction now makes perfect sense.

README.md
```diff
@@ -1680,35 +1683,41 @@ namespace std {
...
-### Class `thread_pool`
+### Class `static_thread_pool`
 
 This class represents a statically sized thread pool as a common/basic resource
-type. This pool provides an effectively unbounded input queue and as such calls
-to add tasks to a thread_pool's executor will not block on the input queue.
+type. This pool is capable of being grown by attaching threads to the pool but
+it will not change size in any automatic way. This pool provides an
+effectively unbounded input queue and as such calls to add tasks to a
+thread_pool's associated executors will not block on the input queue.
```
chriskohlhoff: Missed a `static_thread_pool` here.

ccmysen: agreed.

ccmysen: updated based on the comments, though i updated the wording to indicate that in the static pool you are obliged to use the threads to execute work. I call it required to use the attached threads, we can discuss if we like that wording.

ccmysen:
I think this is ready to merge now.

jaredhoberock:
Thanks Chris.

jaredhoberock:
I can't do it on my end, unfortunately. It ought to be as simple as selecting R1 from the drop-down box at the top of the PR.

jaredhoberock:
Merged in #118.

PR #105, opened 2017-02-03
Rewrite introduction

jaredhoberock:
This pull request reworks the introduction to attempt to provide a better context for the specification which follows.


jaredhoberock:
I'll continue tweaking this, but will merge now so we can see the current state of the paper.

PR #115, opened 2017-02-04
Specify bulk executor customization points

jaredhoberock:
This is a work in progress which ports the bulk customization points from the old style to the new style.


jaredhoberock:
`bulk_then_execute()` falls back to `then_execute()` + `bulk_sync_execute()` when `can_execute_v && can_bulk_sync_execute_v`. Depending on how `can_bulk_sync_execute_v` is specified, this could introduce a cycle.

It's not totally clear to me where best to avoid that cycle -- in the specification of `can_bulk_sync_execute_v` or in the specification of `execution::bulk_then_execute`. Perhaps the latter.

PR #106, opened 2017-02-03
Execution context design discussion

chriskohlhoff:
Work in progress.

PR #120, opened 2017-02-04
Integrate ChrisK's execution context design discussion from #106

jaredhoberock:
This change merges some of the discussion from #106 into the introduction.


ccmysen:
Not sure if this is still work in progress, but i think it's important to call a context a handle which represents and can be used to interact with a set of resources. They are not necessarily about management of the resources as in many cases the context is not actually manage the resources at all.

I also still wonder if an example why knowing if contexts are equal can be useful would help the discussion.

PR #124, opened 2017-02-04
Finish requirements on execution functions

AerialMantis:
* Add requirements for bulk cardinality.
* Add requirements for directionality.
* Add requirements for blocking semantics.
* Add further wording to clarify precedence of semantics from different
properties.

I changed the order of the sections on each property's requirements to have blocking semantics, then directionality and then carnality, although it's different from the naming ordering, it is easier to describe the additions for each property this way.

I have defined a precedence between semantics in order to avoid the need for titles for the semnatics and specifying whether semantics are replaced, this seams to make the wording much simper than in the generalised executors proposal.

I also moved the position of the `pred` parameter for the bulk execution functions so that it is before `n, rf, sf` as this means we can include this is the requirements for the directionality property and removes the need for a corner case rule in the carnality requirements. Does this cause an issue for anyone? If so I can do this another way, this was just simpler. If people are happy with this I can update the order of the parameters else-ware in the paper.

I wasn't sure how we should define the relationship between the execution function name from previous properties and the prefix that is added for a particular property. The way I've done it here is by having `'e'` denote the name of the execution function from previous properties and then the prefix is added to that. For example: the synchronous two-way member execution function would be denoted as `x.sync_'e'(...)`. However I am open to other suggestions for alternative wording for this.

What does everyone think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -563,6 +563,58 @@ The name of an execution function is determined by the combination of properties
 |-------------|----------------|-------------------|
 | "" or "bulk_" | "" or "sync_" or "async_" | "execute" or "post" or "defer" |
 
+#### Blocking semantics
+
+The blocking semantics of an execution function may be one of the following:
+
+* *Potentially blocking:* The execution function may block the caller pending completion of the submitted function objects. Execution functions having potentially blocking semantics are named `execute`.
+* *Non-blocking:* The execution function shall not block the caller pending completion of the submitted function objects. Execution functions having non-blocking semantics are named `post` or `defer`.
+
+##### Requirements on execution functions having potentially blocking semantics
+
+In the Table below, `x` denotes a (possibly const) executor object of type `X`, `e` denotes the name of the execution function, `f` denotes a function object of type `F&&` callable as `DECAY_COPY(std::forward<F>(f))()` and where `decay_t<F>` satisfies the `MoveConstructible` requirements, and `a` denotes a (possibly const) value of type `A` satisfying the `ProtoAllocator` requirements.
+
+| Expression | Return Type | Operational semantics | Assertion/note/ pre-/post-condition |
+|------------|-------------|------------------------|------------------------------------|
+| `x.execute(f, ...)` <br/>`execute(x, f, ...)` | void | *Creation of agents:* Creates a weakly parallel execution agent which invokes `DECAY_COPY( std::forward<F>(f))()` at most once, with the call to `DECAY_COPY` being evaluated in the thread that called `e`. <br/> <br/> *Forward progress of caller:* May block forward progress of the caller until `DECAY_COPY( std::forward<F>(f))()` finishes execution. | *Synchronization:* The invocation of `execute` synchronizes with (C++Std [intro.multithread]) the invocation of `f`. |
+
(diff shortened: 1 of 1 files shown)


jaredhoberock:
> I also moved the position of the pred parameter for the bulk execution functions so that it is before n, rf, sf as this means we can include this is the requirements for the directionality property and removes the need for a corner case rule in the carnality requirements. Does this cause an issue for anyone? If so I can do this another way, this was just simpler. If people are happy with this I can update the order of the parameters else-ware in the paper.

Let's keep the positioning of the `pred` parameter for now. Moving it causes issues with some extensions to the bulk operations we have planned.

README.md
```diff
@@ -559,82 +559,122 @@ An execution function is a member function of the form:
...
+##### Requirements on execution functions of single cardinality
 
-##### Requirements on execution functions having non-blocking semantics
+In the Table below, `x` denotes a (possibly const) executor object of type `X`, `'e'` denotes the name of the execution function from previous properties, `f` denotes a function object of type `F&&` callable as `DECAY_COPY(std::forward<F>(f))()` and where `decay_t<F>` satisfies the `MoveConstructible` requirements, and `a` denotes a (possibly const) value of type `A` satisfying the `ProtoAllocator` requirements.
 
-*TODO*
+| Expression | Return Type | Operational semantics |
+|------------|-------------|---------------------- |
+| `x.'e'(..., a)` | ... | Equivalent to `x.'e'(..., executor_default_allocator_v<X>(x))`. |
+| `'e'(x, ..., a)` | ... | Equivalent to `'e'(x, ..., executor_default_allocator_v<X>(x))`. |
```
chriskohlhoff: I think the `, a` is not supposed to be in the two expressions above.

chriskohlhoff: Also we're currently missing executor_default_allocator from the paper.

I think we don't need to say anything at all right now in the "Operational semantics" of these first two entries. For example I don't think std::function specifies what allocator is used when you call its non-allocator constructor.

We can then look at adding a full specification of executor_default_allocator later.

README.md
```diff
@@ -559,82 +559,122 @@ An execution function is a member function of the form:
...
+| `x.'e'(..., a)` <br/>`'e'(x, ..., a)` | ... | Executor implementations should use the supplied allocator (if any) to allocate any memory required to store the function object. Prior to invoking the function object, the executor shall deallocate any memory allocated. [*Note:* Executors defined in this Technical Specification always use the supplied allocator unless otherwise specified. *--end note*] |
```
chriskohlhoff: Also just noticed that the `(if any)` should be deleted here.

AerialMantis:
@jaredhoberock I've altered the way the execution functions are described so that they take into account the position of the pred parameter, so we can keep the ordering how it is.

@chriskohlhoff I've removed the ", a" from the execution functions and removed the mention of the default allocator parameter from the operational semantics. Since the two rows are now equivalent in terms of operational semantics i've also merged those together so there's a single row for both overloads, i've left the "if any" text in as it now makes sense in this context. Is this what you had in mind?

chriskohlhoff:
Looks good. I'm going to merge this to make newer pull requests cleaner.

PR #126, opened 2017-02-04
Missing bulk customization points

jaredhoberock:


jaredhoberock:
I didn't specify the fallbacks for `bulk_async_post` and `bulk_async_defer` analogously to the fallbacks for `async_post` and `async_defer` because it wasn't immediately clear to me how to do that. We can revisit those fallbacks later if we want to specify them.

PR #127, opened 2017-02-04
Use free function detectors

jaredhoberock:
Here's an idea for reducing some boilerplate in the customization point specifications. I've done it just for `execute()` to show what it would look like.

@chriskohlhoff, does this wording look OK to you?

### DIFF
--- a/README.md
+++ b/README.md
@@ -853,9 +853,11 @@ This sub-clause contains templates that may be used to query the properties of a
 
 ## Customization points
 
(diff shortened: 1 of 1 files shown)


chriskohlhoff:
On Sun, Feb 5, 2017, at 08:41 AM, Jared Hoberock wrote:

> Here's an idea for reducing some boilerplate in the
> customization point
> specifications. I've done it just for `execute()` to show what
> it would
> look like.

>

> @chriskohlhoff, does this wording look OK to you?



I like it.

PR #136, opened 2017-02-05
Add allocator support to execute customizaton point.

chriskohlhoff:
I've tried to find a formulation that adds the optional allocator parameter, but without doubling the size of the customization point specification.

Please let me know if it looks ok and I'll complete the changes for the other customization points where this applies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -907,15 +907,15 @@ The name `execute` denotes a customization point. The effect of the expression `
       constexpr unspecified post = unspecified;
     }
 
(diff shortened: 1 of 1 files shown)


jaredhoberock:
That looks good to me. Is this ready to be merged?

chriskohlhoff:
Yep, if you're ok with the approach then it's ready.

jaredhoberock:
Yep, I'm good with it. Merged.

PR #144, opened 2017-02-05
Customization points graph

chriskohlhoff:
The customization_points.gv file is the input file.

The .snippet file contains a piece of html that can be inserted into an html file. The image data is embedded so that the document remains self contained.

The perl script can be run to regenerate the .png and .snippet files.


jaredhoberock:
Nice, is this ready to merge? It looks like there are some commented-out customization points which don't appear in the diagram yet.

chriskohlhoff:
I looked at the current specification of then_execute and bulk_then_execute and (unless I missed it) they don't seem to adapt any other customization point. Did you want them to appear in the graph as standalone boxes anyway?

jaredhoberock:
Maybe I simply misunderstood what the diagram was visualizing.

The current specification of `bulk_then_execute` will call `then_execute` + `bulk_sync_execute` as its fallback

`then_execute` doesn't call any other customization, but it will call `future.then()` as the fallback.

chriskohlhoff:
I missed that in bulk_then_execute. Will push fix.

jaredhoberock:
Thanks, merged.

PR #150, opened 2017-02-06
Complete R1

jaredhoberock:
I think the paper is ready for submission to Kona.


chriskohlhoff:
I just gave it a very quick eyeball and can't see any obvious problems, except that the footnote should use 'P' and not 'D'.

PR #160, opened 2017-03-31
Restore concepts relating to execution context and resources

hcedwar:


jaredhoberock:
Thanks Carter. I'll incorporate this material into the design document paper.

PR #165, opened 2017-04-07
Changes to explanatory document

AerialMantis:
* Add additional terms to terminology section.
* Add examples to terminology section.
* Minor fixes.
* Add section for ongoing discussions.
* Add example of adaptor type trait.


AerialMantis:
I've finished making the changes on this pull request, I added a paragraph about allocator parameters and finished adding the justifications for the one way execution functions. For `bulk_execute` I put an example of implementing it using a for loop over `execute` at the point of submission to demonstrate why adapting in such a way could be problematic semantically and for performance.

PR #170, opened 2017-04-28
Define bound for execution agents

hcedwar:
Elaborate on what it means for an execution agent to be bound to an execution context.


jaredhoberock:
Thanks Carter

PR #175, opened 2017-05-01
Fixes several spellings and added OpenMP blocking reasons

fraggamuffin:


jaredhoberock:
Thanks Michael, I will turn these footnotes into issues so that they can be tracked.

PR #176, opened 2017-05-01
Add linkage with SG5 TM

fraggamuffin:
I wanted to add how executors interact with SG5 TM in the proposal but ran out of time. We at SG5 is starting to study his interaction, and I wanted to put a placeholder here (actually in the last proposal but I ran out of time)  in case anyone wonders why interaction with a concurrency mechanism like TM is not stated.

### DIFF
--- a/explanatory.md
+++ b/explanatory.md
@@ -466,6 +466,10 @@ and technical specifications.
 
 Table: The control structures we propose to introduce.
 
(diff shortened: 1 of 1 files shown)


fraggamuffin:
Looking deeply at it, I dont think it will be too different then task_blocks.

PR #187, opened 2017-05-12
Reword static_thread_pool::attach.

chriskohlhoff:
- Put requirement that thread will be used for execution up front.

- Remove Windows rationale that is not relevant to a static thread pool (it
  was carried over from when the class still had dynamic thread pool
  capabilities.)


jaredhoberock:
Thanks Chris.

PR #192, opened 2017-05-30
Rename these types related to the bulk forward progress traits

### LINKED ISSUE #163: Consider renaming these two executor type traits
In explaining how the executor type traits work, it occurs to me that there may be an opportunity to improve the naming of a couple of the traits to increase clarity and consistency.

What do people think of these name changes? Are they improvements?

* `executor_execute_blocking_category` -> `executor_execute_blocking_guarantee`
* `executor_execution_category` -> `executor_bulk_ordering_guarantee`

The `category` naming scheme was inspired by `iterator_category`, but the things we're describing aren't really categories, and anyway we use the term "executor category" to describe something completely different. The things we want to communicate with respect to blocking and bulk execution are guarantees.

I'd also consider eliminating the `_tag` suffixes on the possible values for these traits because it's another iterator trait-inspired thing that just seems like noise.

jaredhoberock:
Correct some false statements about post & defer's blocking behavior
Fix some stray merge damage in the wording document

Fixes #163

### DIFF
--- a/explanatory.md
+++ b/explanatory.md
@@ -987,19 +987,19 @@ The default value of `executor_execution_mapping_category` is `thread_execution_
 
 **Blocking guarantee.** When a client uses an executor to create execution
 agents, the execution of that client may be blocked pending the completion of
-those execution agents. The `executor_execute_blocking_category` trait
+those execution agents. The `executor_execute_blocking_guarantee` trait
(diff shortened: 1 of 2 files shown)


jaredhoberock:
Reverted the changes to the names of the blocking traits because pull request #195 does this separately.

Issue #164, opened 2017-04-07
Consider promoting .context() to be a customization point


jaredhoberock:
Our current proposal requires every executor to define a `.context()` member function. Right now, it is the only interface requiring clients to go directly through a member rather than indirectly through a customization point or type trait. Moreover, if an executor type does not have an independent context they are allowed act as their own context, because all executor types trivially fulfill our small list of requirements for execution contexts.

It seems like we could improve the uniformity of our design and further eliminate a small extra requirement of executor authors (the `.context()` member function) if we introduced the customization point `execution::context(executor)`. Its behavior would be simple:

  * If `executor.context()` is well-formed, return its result; otherwise
  * if `context(executor)` is well-formed, return its result; otherwise
  * `return executor`

ccmysen:
The last property would imply then that any copy of the executor would actually not be the same context (e.g. passing an executor to a function by value would then place it in a unique context)?

I think the customization point is consistent, though i'm not entirely sure there is a meaningful way for external code to know how to infer or generate a context, it seems like something that must come from the executor.

I think it's okay to say that it's acceptable for an executor to return itself as the context, but probably undesirable to do so unless no other context exists.

jaredhoberock:
Yes to customization point, no to default. Third bullet should be ill-formed.

jaredhoberock:
Apply the same rationale to the work tracking functions. Third bullet should be ill-formed.

jaredhoberock:
Decided against merging #193 on June 9, 2017 telecon

PR #193, opened 2017-05-30
Introduce execution::context()

jaredhoberock:


jaredhoberock:
I submitted this pull request to introduce the new customization points mentioned in #164. However, after having written the text, and considering the change to `ExecutorWorkTracker`, I'm no longer convinced it's a necessary to introduce these change right now.

Currently, all customization points are execution functions which adapt an executor's native execution functions to provide the semantics of the requested execution function. This is the definition given in S 2.2.

However, the new customization points proposed in this issue do not adapt executor behavior. Uses of these customization points become compile-time errors when the executor does not natively provide the requested behavior. The purpose of customization points is to enable uniform use of executors in generic contexts, and these changes seem to defeat that purpose.

I also encountered an issue when attempting to produce wording which would introduce customization points for the work tracking functions `on_work_started` and `on_work_finished`. As far as I can tell, these functions do not take executors per se as parameters -- they take `ExecutorWorkTracker` objects. Currently, all customization points apply to executors. It wasn't clear to me if we intended to open up the customization points to types which are not strictly executors. Moreover, these work tracking functions still seem like hints to me. It wasn't obvious to me if they have any semantics.

Anyway, maybe we shouldn't introduce these customization points right now. We can always add them later if we are convinced they are necessary.

jaredhoberock:
Decided against merging #193 on June 9, 2017 telecon

PR #196, opened 2017-05-30
Change occurrences of std::future to std::experimental::future

### LINKED ISSUE #188: Should executor_future default to std::experimental::future?
P0443R1 2.3.3 specifies that `executor_future<Executor,T>` defaults to `std::future<T>` when `Executor` has no execution functions that produce futures:

> ```
> template<class Executor, class T>
> struct executor_future
> {
>   using type = see below;
> };
> ```
> 
> The type of `executor_future<Executor, T>::type` is determined as follows:
> 
> * if `is_two_way_executor<Executor>` is true, `decltype(declval<const Executor&>().async_execute( declval<T(*)()>());`
> 
> * otherwise, if `is_one_way_executor<Executor>` is true, `std::future<T>;`
> 
> * otherwise, the program is ill formed.

However, P0443R1 2.2.1 `Future` requirements says:

> A type `F` meets the `Future` requirements for some value type `T` if `F` is `std::experimental::future<T>` (defined in the C++ Concurrency TS, ISO/IEC TS 19571:2016). [Note: This concept is included as a placeholder to be elaborated, with the expectation that the elaborated requirements for `Future` will expand the applicability of some executor customization points. --end note]

I believe this discrepancy is simply an oversight, and that `executor_future`'s specification should be updated to default to `std::experimental::future<T>` instead of `std::future<T>`.

jaredhoberock:
Change occurrences of std::shared_future to std::experimental::shared_future

Fixes #188

### DIFF
--- a/wording.md
+++ b/wording.md
@@ -692,7 +692,7 @@ The name `async_execute` denotes a customization point. The effect of the expres
 
 * Otherwise, `async_execute(E, F, A...)` if `has_async_execute_free_function_v<decay_t<decltype(E)>>` is true.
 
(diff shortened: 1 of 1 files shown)


jaredhoberock:
This was a very simple change, so I'm going to go ahead and merge it now to get it off the list of open pull requests.

PR #197, opened 2017-06-01
sync_ functions return Futures and one-way functions shall not throw user exceptions

jaredhoberock:
These changes update `sync_` functions to return `executor_future_t` and also ban user exceptions from escaping one-way functions.

### DIFF
--- a/explanatory.md
+++ b/explanatory.md
@@ -536,18 +536,19 @@ create multiple agents in bulk have names prefixed with `bulk_`;
 single-agent customization points have no prefix. 
 
 **Directionality.** Some executor customization points return a `Future` object
-for synchronizing with and retrieving the result or exception of execution. Other
-customization points allow clients to "fire-and-forget" their execution and
-return no `Future`. We refer to fire-and-forgetful customization points as
+for synchronizing with and retrieving the result or exception of execution.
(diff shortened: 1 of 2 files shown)


explanatory.md
```diff
@@ -1830,20 +1830,6 @@ primitives of full-fledged `std::future` objects. We envision that a
 separate effort will propose a `Future` concept which would introduce
 requirements for these user-defined `std::future`-like types.
 
-### Error Handling
```
AerialMantis: Should we have something here that describes the ways in which an implementation could provide a communication mechanism for user exceptions?  Something to the effect of:

"The proposal prescribes no mechanism for functions to communicate exceptional behavior back to their clients. However the interface is designed to permit an implementation to propagate user exceptions from two-way execution functions via the returned future object and to permit an implementation to support communication of exceptions from one-way execution functions via an back channel."

jaredhoberock: Our design requires two-way executors to return exceptions through the `Future`s they return, doesn't it? Or is exception handling executor-defined in all cases?

Also, I believe there is a note about backchannels earlier in the document. The idea of this change was to completely eliminate this future work section on error handling.

PR #199, opened 2017-06-09
Change discussing default constructor for static_thread_pool

### LINKED ISSUE #158: Include a default constructor to thread pool
Preference from committee to have a default thread pool constructor which does something implementation defined (allowing for a default "2x num cores" or something similar).

This _might_ motivate having a way to query the number of threads currently joined to the pool in case you need to join more to it.

ccmysen:
Resolves #158 by adding wording around the rationale for no default constructor for the static_thread_pool.


jaredhoberock:
Thanks Chris.

PR #254, opened 2017-08-13
Issue 241: Add default properties for thread mapping and allocator.

AerialMantis:
* Add `default_execution_mapping` property.
* Add `default_allocator` property.
* Add wording to describe directionality and cardinality properties as
accumulative.
* Add wording to describe allocator property as mutually exclusive.
* Rename `default_execution_mapping` -> `other_execution_mapping`.
* Re-order all properties to be in order of guarantees.
* Add wording which describes the relative ordering of property
guarantee subsets.

### DIFF
--- a/wording.md
+++ b/wording.md
@@ -47,7 +47,7 @@ namespace execution {
 
   // Memory allocation properties:
 
-  struct default_allocator_t {} deault_allocator;
+  struct default_allocator_t {} default_allocator;
   template<class ProtoAllocator> struct allocator_t { ProtoAllocator alloc; };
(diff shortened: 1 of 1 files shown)


ccmysen:
BTW -- there are a couple options for defaulted /optional values.
There was the status & value that was quoted in the call, there's also this one:
http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/p0323r2.pdf

And std::optional

Optional is the simplest and communicates whether a value exists or not and that may be sufficient. status_and_optional is an extension which allows for communicating more information than a boolean has_value.

AerialMantis:
Thanks @ccmysen, I've created a merge request for the rest of the changes for query, and I've included examples using `std::optional`, `std::expected` and `std::status_value` in the explanatory document.

PR #259, opened 2017-09-08
Issue 244: Add query customisation point

AerialMantis:
Add a new customisation point called `query` which dynamically returns the value of a property or other information.


AerialMantis:
@jaredhoberock this should be ready to merge in now.

jaredhoberock:
Thanks Gordon!

PR #269, opened 2017-10-14
Changelog & metadata updates

jaredhoberock:
Since we added support for guaranteeing forward progress wrt the caller in R3, I think we need no longer suggest that this is an open design issue.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,10 +5,10 @@ wording: wording_pdf wording_html
 explanatory: explanatory_pdf explanatory_html
 
 wording_pdf: wording_front_matter.md wording.md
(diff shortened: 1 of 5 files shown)


jaredhoberock:
Updated the paper metadata & added a changelog.

jaredhoberock:
@chriskohlhoff @AerialMantis @ccmysen, if you see this before tomorrow morning, take a quick look at the changelog and let me know if there are items I've missed that ought to be mentioned.

chriskohlhoff:
I'd suggest changing:

> Introduced properties allowing executors to opt into one-way to two-way adaptations

to something like:

> Introduced properties allowing executors to opt into adaptations that add blocking semantics

jaredhoberock:
Thanks, I've made that change and merged.

PR #271, opened 2017-10-15
Add discussion of alternative names for require and prefer.

chriskohlhoff:


chriskohlhoff:
@AerialMantis can you please confirm that this concurs with the feedback at cppcon. Once you're happy with it then it should be ready for @jaredhoberock to merge.

AerialMantis:
@chriskohlhoff yeah this looks good, there was also feedback that `transform` could conflict with the algorithm, so we could mention that as a consideration as well.

chriskohlhoff:
Done. @jaredhoberock this one is ready to merge if you're happy with it.

jaredhoberock:
LGTM. Thanks.

PR #313, opened 2018-01-05
Context as property

chriskohlhoff:
Work-in-progress pull request. Replaces context() member function with a query-only property.


jaredhoberock:
I accidentally merged this into R4 when it should have gone into R5. I've reverted the change, but it doesn't look like there's a way to resubmit the PR pointed at R5.

@chriskohlhoff: Could you resubmit this pull request?

PR #318, opened 2018-01-12
Acknowledge the issue of interoperation between executors & coroutines

### LINKED ISSUE #302: Acknowledge the issue of interoperation with coroutines
It will be sufficient to acknowledge that there must be a story for integration between executors and coroutines in the explanatory document, but this story is not required for an Executor TS v1.

jaredhoberock:
This change adds a paragraph to the envisioned extensions section of the explanatory document mentioning the issue of interactions between executors and coroutines and also bumps the explanatory document's revision number.

Fixes #302

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,10 +17,10 @@ CITEPROC= --filter pandoc-citeproc \
 	  --csl=acm-sig-proceedings-long-author-list.csl
 
 explanatory_pdf: explanatory.md explanatory_header.tex explanatory_metadata.yaml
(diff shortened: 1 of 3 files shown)


chriskohlhoff:
We could add a reference to P0286R0 with a note to the effect: based on prior experience with a similar executor model we believe that interoperation is possible, although the exact form is still to be determined.

jaredhoberock:
Thanks, I've incorporated that suggestion and cited P0286.

PR #321, opened 2018-01-29
New properties model where nested constants/types define property characteristics

chriskohlhoff:
Work in progress.


wording.md
```diff
@@ -356,17 +364,33 @@ The `oneway`, `twoway` and `then` properties are not mutually exclusive.
...
+    struct S
+    {
+      static constexpr bool is_requirable = true;
+      static constexpr bool is_preferable = true;
+
+      using polymorphic_query_result_type = bool;
+
+      template<class Executor>
+        static constexpr bool is_supportable
+          = can_query_v<Execution, S>;
```
jaredhoberock: Typo: Execution -> Executor

jaredhoberock:
I think this looks good so far.

@chriskohlhoff: What do you think about an additional step that would refactor `execution::require`'s specification? For each of the "standard adaptations", we could introduce an overload of `require`.

E.g., for `always_blocking_t`:

    namespace execution
    {

    struct always_blocking_t
    {
      ...

      template<class Executor>
      friend see-below require(const Executor& ex, const always_blocking_t&);
    };

And move the documentation of the always blocking adaptation into the specification `always_blocking_t`'s `require` overload. I believe the use of a `friend` function can avoid a collision with `execution::require`.

chriskohlhoff:
Yes let's do that. However, I have a feeling that in general the library standardese avoids the explicit use of `friend` and instead prefers to simply specify them as associated functions. We might need to rework this later.

N.B. maybe it should be:

    template<class Executor>
      friend see-below require(Executor ex, const always_blocking_t&);

to allow temporaries to be moved from.

chriskohlhoff:
Also once we've done this we should probably reorder the paper so that the customization points come before the properties. Let's do that last to keep the diffs sane.

jaredhoberock:
Sounds good.

I think a lot of properties may end up having this form:

    struct S
    {
      static constexpr bool is_preferable = true;

      using polymorphic_query_result_type = bool;

      template<class Executor>
        static constexpr bool is_supportable
          = can_query_v<Executor, S>;
    };

Would it be useful to create some resuable shorthand for that form?

chriskohlhoff:
I think we can do that by hoisting the `struct S` form to the top of the behavioral properties section. We'll need an exception for allocator, but I think that's the only one. I'll push a commit with this shortly.

chriskohlhoff:
The other exceptions would of course be properties with adapters. We might be able to address those with the following form.

In addition to conforming to the form above, the `always_blocking_t` has the following specification:

    struct always_blocking_t
    {
      template<class Executor>
      friend see-below require(Executor ex, const always_blocking_t&);
    };

etc.

AerialMantis:
This looks great so far!

I agree that it would be good if we could have a generic form which could be used for all the properties (except the exceptions). We could perhaps put placeholders for the values of `is_requireable`, `is_preferable`, `polymorphic_query_result_type ` and `is_supportable` and have them listed in a table.

I've also been thinking about how we could define the form for properties which have a value and are require-able and query-able like the `allocator` property. I think we should have a general form which can be used for any properties like this, it would make the design more future proof.

I sketched up a potential form based on what we have for the `allocator` property at the moment:

```cpp
template <typename T>
struct S
{
    static constexpr bool is_preferable = true;
    using polymorphic_query_result_type = bool;
    using value_type = some_type;

    template<class Executor>
    static constexpr bool is_supportable = can_query_v<Executor, S<T>>;

    template <typename D = T, typename = typename std::enable_if_t<std::is_same_v<D, T> && std::is_same_v<T, void>>>
    S<value_type> operator()(const value_type &value) const { return S<value_type>{}; }
};

constexpr S<void> s;
```

The main addition for this form is the `value_type` alias specifying the type of the value associated with the property and a function call operator (which is only available to the void specialisation, and which takes `const value_type &`. This is not ideal so perhaps we want may want to think of alternative solutions, such as having a specific type defined for containing property-value pairs, though this does allow us to use the same name for both uses of the property.

AerialMantis:
Also, let me know if there is anything I can do to help with the wording, I don't want to conflict with any changes your working on.

jaredhoberock:
As of this message, it looks like the properties remaining to be ported are `context` and `allocator`. I will let either @AerialMantis or @chriskohlhoff take care of those since it seems like you guys might have some ideas for how to handle them.

chriskohlhoff:
`context` is done, so that just leaves allocator.

@AerialMantis some random thoughts about how the allocator_t property should look...

Do we want it to work with the polymorphic executor? If not (i.e. the much simpler option) then we can drop the polymorphic_query_result_type typedef.

Do we want to say that allocator-aware executors are always queryable with allocator_t<void>? This would allow users to ask an executor for its allocator even if they don't the concrete type, i.e. seems like something we'd want. If so, how about we enforce it by making `is_supportable` be `can_query_v<Executor, allocator_t<T>> || can_query_v<Executor, allocator_t<void>>`.

Shouldn't the argument to operator() be a template parameter (i.e. ProtoAllocator)? I also think we could keep operator() even in the non-void case. It's conceptually like a rebind.

AerialMantis:
@chriskohlhoff @jaredhoberock I can do the wording for allocator.

> Do we want it to work with the polymorphic executor? If not (i.e. the much simpler option) then we can drop the polymorphic_query_result_type typedef.

I think we should still aim for the allocator property to work with the polymorphic executor if possible. Though if the property is templated then `polymorphic_query_result_type` could potentially differ per allocator type as a specialisation of the property, but is this something we would want? You would have to use the specialisation of the property to derive the polymorphic query result type `allocator_t<my_allocator>::polymorphic_query_result_type`. I would suggest that we require all specialisations of the property to provide the same type for `polymorphic_query_result_type` or we require that only the `void` specialisation provide the traits and the `T` specialisations be only for storing the value.

Though this raises another question, what type is used in the polymorphic executor property list, I imagine it should be `allocator_t<void>`, and this would allow any allocator_t<T> type.

> Do we want to say that allocator-aware executors are always queryable with allocator_t<void>? This would allow users to ask an executor for its allocator even if they don't the concrete type, i.e. seems like something we'd want. If so, how about we enforce it by making is_supportable be can_query_v<Executor, allocator_t<T>> || can_query_v<Executor, allocator_t<void>>.

Yes, I think we should make `is_supportable` require `can_query_v<void>`, though i'm not sure if we need to require `can_query_v<T>`, would this result of this be a boolean? Alternatively, you could call `query(allocator)` and compare the result.

> Shouldn't the argument to operator() be a template parameter (i.e. ProtoAllocator)? I also think we could keep operator() even in the non-void case. It's conceptually like a rebind.

Yes you're correct, the function call operator should take the allocator type as a template parameter.

jaredhoberock:
Re: `allocator_t::polymorphic_query_result_type`: is `std::pmr::polymorphic_allocator` useful here?

chriskohlhoff:
It looks like `pmr::polymorphic_allocator` can only be used to wrap a `memory_resource`, and can't type-erase arbitrary allocators. The other data point to consider is that `std::function` has had its allocator support removed, so we may be on firmer ground omitting allocator support from the polymorphic wrapper than trying to put it in.

jaredhoberock:
Fair enough; I had assumed it was a type eraser for allocators. It seems to me that's what we really require here, but it's not our job to design one. I'd suggest omitting support for allocator support for now; it can always be introduced later if we get an actual allocator type eraser.

AerialMantis:
I agree, unfortunately, I don't think the polymorphic allocator would help here, as it only allows for wrapping memory resources.

I think it's fair to omit allocator support for the polymorphic executor for now, but continue to look into supporting it in the future, as we may at some point introduce similar properties which we need to support in the polymorphic executor. If we decide to omit support for the current paper I'd be happy to look into this for the following paper.

chriskohlhoff:
Ready for merging. As discussed, @AerialMantis you could use a separate pull request for your allocator properties chanages.

jaredhoberock:
Thanks, merged.

ccmysen:
Sent a mail about this as well:

- For adaptable_blocking, there's not really a way to tell if adaptation has occurred, and in fact the conversion to two-way would fail if adaptation was not allowed (but succeeds if it is). How are we thinking about that expression in the properties and/or how require works. This is basically a rehash of the dependencies question, but still highlights the issue of adaptation being dependent on other properties existing in order for adaptation to succeed. Minimally we would probably need to require adaptable_blocking to be required before the conversion to two_way in the documentation, though it does mean depending on ordering you may succeed or fail to compile.
- I thought we were going to drop the ordering guarantees for bulk submission (was that just something that we still have because we didn't remove it yet)?
- Our wordings are stated to reference "an executor's execution function" whereas an executor may have multiple execution functions. This is there in the wording already, but we probably should be more explicit about which execution functions are affected to clarify that properties apply to which parts of the interface. We already do this explicitly in the requirements when interface-affecting properties interact (if you request twoway_t it must satisfy TwoWayExecutor or BulkTwoWayExecutor and if you request bulk_t it must satisfy BulkOneWayExecutor, BulkTwoWayExecutor or BulkTheExecutor. The implication is that twoway_t + bulk_t must at least satisfy BulkTwoWayExecutor. I don't know that we actually state that BulkTwoWayExecutor must be satisfied though. Certainly with behavioral properties we should be clear which parts of the interface are affected.
- It's not obvious why a query only property for context_t would have is_requireable = true? is that a mistake?

PR #325, opened 2018-02-04
Property hoisting

chriskohlhoff:
Work in progress. Needs review.

### DIFF
--- a/wording.md
+++ b/wording.md
@@ -654,6 +654,27 @@ template<class Executor>
 
 *Remarks:* This function shall not participate in overload resolution unless `can_query_v<Executor, always_blocking_t> || can_query_v<Executor, never_blocking_t>` is false.
 
+##### `always_blocking_t` customization points
+
+In addition to conforming to the above specification, the `always_blocking_t` property provides the following customization:
+
+    struct always_blocking_t
+    {
+      template<class Executor>
+        friend see-below require(Executor ex, always_blocking_t);
+    };
+
+If the executor has the `adaptable_blocking_t` property, this customization uses an adapter to implement the `always_blocking_t` property.
+
+```
+template<class Executor>
+  friend see-below require(Executor ex, always_blocking_t);
+```
+
+*Returns:* A value `e1` of type `E1` that holds a copy of `ex`. If `Executor` satisfies the `OneWayExecutor` requirements, `E1` shall satisfy the `OneWayExecutor` requirements by providing member functions `require`, `query`, and `execute` that forward to the corresponding member functions of the copy of `ex`. If `Executor` satisfies the `TwoWayExecutor` requirements, `E1` shall satisfy the `TwoWayExecutor` requirements by providing member functions `require`, `query`, and `twoway_execute` that forward to the corresponding member functions of the copy of `ex`. If `Executor` satisfies the `BulkOneWayExecutor` requirements, `E1` shall satisfy the `BulkOneWayExecutor` requirements by providing member functions `require`, `query`, and `bulk_execute` that forward to the corresponding member functions of the copy of `ex`. If `Executor` satisfies the `BulkTwoWayExecutor` requirements, `E1` shall satisfy the `BulkTwoWayExecutor` requirements by providing member functions `require`, `query`, and `bulk_twoway_execute` that forward to the corresponding member functions of the copy of `ex`. In addition, `E1` provides an overload of `require` such that `E1.require(always_blocking)` returns a copy of `E1`, an overload of `query` such that `E1.query(always_blocking)` returns `true`, and all functions `execute`, `twoway_execute`, `bulk_execute`, and `bulk_twoway_execute` shall block the calling thread until the submitted functions have finished execution. `e1` has the same executor properties as `ex`, except for the addition of the `always_blocking_t` property, and removal of `never_blocking_t` and `possibly_blocking_t` properties if present.
+
+*Remarks:* This function shall not participate in overload resolution unless `can_query_v<Executor, adaptable_blocking_t>` is true.
+
 #### Properties to indicate if blocking and directionality may be adapted
 
     struct adaptable_blocking_t;
@@ -886,8 +907,6 @@ The name `require` denotes a customization point. The effect of the expression `
 * `(E).require(P0)` if `N == 0` and `has_require_member_v<decay_t<decltype(E)>, decltype(P0)>` is true.
 
 * Otherwise, `require(E, P0)` if `N == 0` and the expression is well formed.
-
-* Otherwise, a value `E1` of unspecified type that holds a copy of `E` if `N == 0`, `is_same<decay_t<decltype(P0)>, always_blocking_t>` is true, and `can_query_v<decltype(E), adaptable_blocking_t>` is true. If `E` satisfies the `OneWayExecutor` requirements, `E1` shall satisfy the `OneWayExecutor` requirements by providing member functions `require`, `query`, and `execute` that forward to the corresponding member functions of the copy of `E`. If `E` satisfies the `TwoWayExecutor` requirements, `E1` shall satisfy the `TwoWayExecutor` requirements by providing member functions `require`, `query`, and `twoway_execute` that forward to the corresponding member functions of the copy of `E`. If `E` satisfies the `BulkOneWayExecutor` requirements, `E1` shall satisfy the `BulkOneWayExecutor` requirements by providing member functions `require`, `query`, and `bulk_execute` that forward to the corresponding member functions of the copy of `E`. If `E` satisfies the `BulkTwoWayExecutor` requirements, `E1` shall satisfy the `BulkTwoWayExecutor` requirements by providing member functions `require`, `query`, and `bulk_twoway_execute` that forward to the corresponding member functions of the copy of `E`. In addition, `E1` provides an overload of `require` such that `E1.require(always_blocking)` returns a copy of `E1`, and all functions `execute`, `twoway_execute`, `bulk_execute`, and `bulk_twoway_execute` shall block the calling thread until the submitted functions have finished execution. `E1` has the same executor properties as `E`, except for the addition of the `always_blocking_t` property, and removal of `never_blocking_t` and `possibly_blocking_t` properties if present.
 
 * Otherwise, a value `E1` of unspecified type that holds a copy of `E` if `N == 0` and `is_same<decay_t<decltype(P0)>, adaptable_blocking_t>` is true. If `E` satisfies the `OneWayExecutor` requirements, `E1` shall satisfy the `OneWayExecutor` requirements by providing member functions `require`, `query`, and `execute` that forward to the corresponding member functions of the copy of `E`. If `E` satisfies the `TwoWayExecutor` requirements, `E1` shall satisfy the `TwoWayExecutor` requirements by providing member functions `require`, `query`, and `twoway_execute` that forward to the corresponding member functions of the copy of `E`. If `E` satisfies the `BulkOneWayExecutor` requirements, `E1` shall satisfy the `BulkOneWayExecutor` requirements by providing member functions `require`, `query`, and `bulk_execute` that forward to the corresponding member functions of the copy of `E`. If `E` satisfies the `BulkTwoWayExecutor` requirements, `E1` shall satisfy the `BulkTwoWayExecutor` requirements by providing member functions `require`, `query`, and `bulk_twoway_execute` that forward to the corresponding member functions of the copy of `E`. In addition, `can_query_v<decltype(E1), adaptable_blocking_t>` is true, and `E1.require(not_adaptable_blocking)` yields a copy of `E`. `E1` has the same executor properties as `E`, except for the addition of the `adaptable_blocking_t` property.
 
--- a/wording.md
+++ b/wording.md
@@ -621,6 +621,39 @@ The `possibly_blocking_t`, `always_blocking_t` and `never_blocking_t` properties
 
 [*Note:* The guarantees of `possibly_blocking_t`, `always_blocking_t` and `never_blocking_t` implies the relationships: `possibly_blocking < always_blocking` and `possibly_blocking < never_blocking` *--end note*]
 
+##### `possibly_blocking_t` customization points
+
+In addition to conforming to the above specification, the `possibly_blocking_t` property provides the following customizations:
+
+    struct possibly_blocking_t
+    {
+      template<class Executor>
+        friend Executor require(Executor ex, possibly_blocking_t);
+
+      template<class Executor>
+        friend constexpr bool query(const Executor& ex, possibly_blocking_t);
+    };
+
+These customizations automatically enable the `possibly_blocking_t` property for all executors that do not otherwise support the `always_blocking_t` or `never_blocking_t` properties. [*Note:* That is, all executors are treated as possibly blocking unless otherwise specified. *--end note*]
+
+```
+template<class Executor>
+  friend Executor require(Executor ex, possibly_blocking_t);
+```
+
+*Returns:* `std::move(ex)`.
+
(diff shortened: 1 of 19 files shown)


wording.md
```diff
@@ -403,6 +383,85 @@ The directionality properties conform to the following specification:
...
+These customizations automatically enable the `oneway_t` property if an executor type already satisfies the syntactic requirements for a oneway executor.
+
+```
+template<class Executor>
+  friend Executor require(Executor ex, oneway_t);
+```
+
+*Returns:* `std::move(ex)`.
+
+*Remarks:* This function shall not participate in overload resolution unless `is_oneway_executor_v<Executor> || is_bulk_oneway_executor_v<Executor>` is true.
```
jaredhoberock: I wonder if we could reword the *Remarks:* clause in such a way to make `is_oneway_executor_v` and `is_bulk_oneway_executor_v` superfluous (and therefore eliminate these traits).

For example:

*Remarks:* This function shall not participate in overload resolution unless `Executor` fulfills syntactic requirements for `OneWayExecutor` or `BulkOneWayExecutor`.

wording.md
```diff
@@ -403,6 +383,85 @@ The directionality properties conform to the following specification:
...
+
+##### `twoway_t` customization points
+
+In addition to conforming to the above specification, the `twoway_t` property provides the following customizations:
+
+    struct twoway_t
+    {
+      template<class Executor>
+        friend Executor require(Executor ex, twoway_t);
```
jaredhoberock: I wonder if it's possible to specify `execution::require()` in such a way as to make these identity `friend` functions unnecessary?

wording.md
```diff
@@ -403,6 +383,85 @@ The directionality properties conform to the following specification:
...
+
+```
+template<class Executor>
+  friend see-below require(Executor ex, twoway_t);
+```
+
+*Returns:* A value `e1` of type `E1` that holds a copy of `ex`. If `Executor` satisfies the `OneWayExecutor` requirements, `E1` shall satisfy the `OneWayExecutor` requirements by providing member functions `require`, `query`, and `execute` that forward to the corresponding member functions of the copy of `ex`, if present, and `E1` shall satisfy the `TwoWayExecutor` requirements by implementing `twoway_execute` in terms of `execute`. Similarly, if `Executor` satisfies the `BulkOneWayExecutor` requirements, `E1` shall satisfy the `BulkOneWayExecutor` requirements by providing member functions `require`, `query`, and `bulk_execute` that forward to the corresponding member functions of the copy of `ex`, if present, and `E1` shall satisfy the `BulkTwoWayExecutor` requirements by implementing `bulk_twoway_execute` in terms of `bulk_execute`. For some type `T`, the type yielded by `executor_future_t<E1, T>` is `std::experimental::future<T>`. `e1` has the same executor properties as `ex`, except for the addition of the `twoway_t` property.
```
jaredhoberock: > For some type `T`, the type yielded by `executor_future_t<E1, T`>` is `std::experimental::future<T>`.

Currently all `Future`s must be `std::experimental::future`. Is this currently redundant statement included for forward portability?

wording.md
```diff
@@ -403,6 +383,85 @@ The directionality properties conform to the following specification:
...
 The `oneway_t`, `twoway_t` and `then_t` properties are not mutually exclusive.
 
+##### `oneway_t` customization points
+
+In addition to conforming to the above specification, the `oneway_t` property provides the following customizations:
+
+    struct oneway_t
+    {
+      template<class Executor>
+        friend Executor require(Executor ex, oneway_t);
```
jaredhoberock: Why do these `require` overloads receive `ex` by value? I would have expected a `const` or forwarding reference, like `query`.

jaredhoberock:
This looks like a big improvement to me.

One question I had while reviewing is how is the `can_query<E,P>` trait related to `P::is_supported<E>`? Aren't they sort of answering similar questions? If so, maybe a better name for `::is_supported` would be something like `::is_queryable`.

jaredhoberock:
`can_query<E,P>::value` is defined to be something like `true` if `execution::query(ex, prop)` is well-formed, and `false` otherwise. This seems like the right definition to me.

However, one issue I see is that `can_query` is used in the definition of properties, which are defined much earlier in the specification.

For example, `context_t`:

```
struct context_t
{

  using polymorphic_query_result_type = any; // TODO: alternatively consider void*, or simply omitting the type.

  template<class Executor>
    static constexpr bool is_supportable
      = can_query_v<Executor, context_t>;
        // TODO consider requiring the query result to be an lvalue reference.
};
```

This arrangement seems kind of circular to me.

One way to break the cycle would be to define `::is_supportable` in terms of whether or not a member `ex.query(p)` or free function `query(ex, p)` exists independently of `execution::query`.

Alternatively, we could define `can_query` earlier, before defining the properties, and without reference to `execution::query()`.

wording.md
chriskohlhoff: I considered this when writing it up (and implementing it in the prototype) and I think it is not possible with the current properties form, unless we can:

* compile-time detect if some query expression is constexpr; and
* standardise what query results mean for presence.

This is because the identity operation, which applies to both Executor type and value, is only valid for properties that are testably present (or not present) at compile time. This is true for the directionality and cardinality properties.

I also suspect that having an identify friend `require` might the simplest expression of this identity characteristic of a property. It makes the property implementer think about whether it applies to their property -- it likely won't -- and keeps the implementation local. Happy for it be proven otherwise, though.

jaredhoberock: > * compile-time detect if some query expression is constexpr; and

I'll attach an example program demonstrating how I implemented it in my prototype, with the caveat that it may just happen to have worked with my compiler.

jaredhoberock:
Here's a program demonstrating how to detect whether a `query` expression is `constexpr`:

```
#include <type_traits>
#include <iostream>

template<class Executor, class Property>
struct is_query_constexpr
{
  template<class E, class P,
           int = (E().query(P()), 0)>
  static std::true_type test(int);

  template<class, class>
  static std::false_type test(...);

  static const bool value = decltype(is_query_constexpr::test<Executor,Property>(0))::value;
};


struct property_t {};

constexpr property_t property{};


struct has_constexpr_query
{
  constexpr bool query(property_t) const
  {
    return true;
  }
};

struct hasnt_constexpr_query
{
  bool query(property_t) const
  {
    return true;
  }
};


int main()
{
  static_assert(is_query_constexpr<has_constexpr_query, property_t>::value, "Error.");
  static_assert(!is_query_constexpr<hasnt_constexpr_query, property_t>::value, "Error.");
  
  std::cout << "OK" << std::endl;

  return 0;
}
```

Output & compiler details:

```
$ clang -std=c++11 is_query_constexpr.cpp -lstdc++
$ ./a.out 
OK
$ clang --version
clang version 3.8.0-2ubuntu4 (tags/RELEASE_380/final)
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```

chriskohlhoff:
> One question I had while reviewing is how is the can_query<E,P> trait related to P::is_supported<E>? Aren't they sort of answering similar questions? If so, maybe a better name for ::is_supported would be something like ::is_queryable.

Not exactly. `is_supportable` is more like "does the executor satisfy the syntactic requirements for this property?". The `can_query` test is (currently) only a subset of that. Though if `can_query_v<E, P>` is a universal requirement of supporting a property, we might consider removing it from `is_supportable` and making it a postcondition of the `require`customisation point.

> This arrangement seems kind of circular to me.
>
> One way to break the cycle would be to define ::is_supportable in terms of whether or not a member ex.query(p) or free function query(ex, p) exists independently of execution::query.

Hmm, I'm not sure I think it's circular exactly, because the customisation points should work only in terms of the Property requirements (which we still have to update btw). I do think we should move the customisation points and their associated traits so that they come first.

chriskohlhoff:
Assuming this `constexpr` test is well-formed (seems to work on my compilers too), maybe we should have a separate property member called `is_present` rather than trying to do something with special `query` result values.

chriskohlhoff:
Please check https://github.com/chriskohlhoff/issaquah_2016/commit/3d8f4e10d72bf2d259aa0190d9ede1de497c8455 (which is currently on a branch off this pull request).

Separating `is_present` from `query` looks like a win with a lot of simplification. The `is_present()` function is like what you suggested on the last call, where the property itself has to "know" what it means for it to be present in an executor. Where it makes sense, a property's `is_present()` can be implemented in terms of `query()`, as the behavioural properties do. Other properties can be present at compile time (like directionality and cardinality) and a `query` function is no longer required for these, i.e. `query` goes back to being for things that *can* be varied at compile time

jaredhoberock:
One possible issue with `is_present()` is how to get a compile-time value for the executor to test. I don't believe something like `std::declval()` works here, because it may only be used in an unevaluated context. Is this actually a problem?

wording.md
AerialMantis: I may be wrong but I remember during the last call discussing the idea of replacing traits such as `is_oneway_executor_v` with `oneway::is_supportable`.

chriskohlhoff:
I think maybe we can make it work if we define `is_present` to take a pointer-to-executor rather than a reference. This lets us use `nullptr`, which is a constant expression, rather than `std::declval`. E.g.:

    struct property_t
    {
      template <class Executor>
      constexpr bool is_present(const Executor*) const { return true; }
    };

    struct executor
    {
    };

    template <class Property, class Executor>
    struct is_compile_time_present
    {
      template <class P, class E> static constexpr auto test(int)
        -> char(&)[P().is_present(static_cast<E*>(nullptr)) ? 2 : 1];

      template <class, class> static char test(...);

      static const bool value = sizeof(test<Property, Executor>(0)) != 1;
    };

    int main()
    {
      std::cout << is_compile_time_present<property_t, executor>::value << "\n";
    }

Then we would simply enable the identity transformation in `require` if `P().is_present(static_cast<decay_t<decltype(E)>*>(nullptr))` is well formed and a constant expression with value `true`.

jaredhoberock:
I'm worried that using a `nullptr` wouldn't pass code review. For example, one issue I see is that `is_present` might depend on the compile-time state of the executor argument.

I admit I'm still having a bit of trouble wrapping my head around the need for both `is_present` and `query`. Do we have a simple example which illustrates how both are used? For example, when is `is_present` different from `can_query_v`?

chriskohlhoff:
`is_present` is different to (i.e. more than) `can_query_v` for all behavioural properties. I.e. for them the well-formedness of `query` is not enough; it's also whether the `query` returns `true`.

If we get back to basics, i would define presence as follows: A property is "present" if-and-only-if the require operation is an identity transformation. So:

* If `P` is present in `E` then `require(E, P) == E`.
* If `require(E, P) == E` then `P` is present in `E`.

In our previous paper revision, `operator==` was the only method available to test for presence. It is sufficient in theory, but possibly not so in practice. Our motivation for introducing some kind of `is_present` operation is twofold:

1. To allow users to test for presence without performing a `require` operation. (So that users may avoid the potentially expensive `require` operation.)
2. To allow the `require` customisation point to automatically perform an identity operation if presence can be determined by other means. (To reduce our specification complexity and alleviate the burden on 3rd party property and executor implementers.)

In some cases, like our interface-changing properties, the presence of a property can be detected at compile time by checking whether syntactic requirements are satisfied. In others, like the behavioural properties, we may not be able to detect presence until runtime. For example:

    class my_executor
    {
      bool is_possibly_blocking = true;

    private:
      bool operator==(const my_executor& rhs) const
        { return is_possibly_blocking == rhs.is_possibly_blocking; }
      bool operator!=(const my_executor& rhs) const
        { return is_possibly_blocking != rhs.is_possibly_blocking; }

      template<class F> void execute(F f) const
      {
        if (is_possibly_blocking)
          try { f(); } catch (...) { std::terminate(); }
        else
          run_in_background(std::move(f));
      }

      my_executor require(execution::possibly_blocking_t) const
      {
        if (is_possibly_blocking)
          return *this; // Already present, identity operation.
        my_executor new_exec;
        new_exec.is_possibly_blocking = true;
        return new_exec;
      }

      my_executor require(execution::never_blocking_t) const
      {
        if (!is_possibly_blocking)
          return *this; // Already present, identity operation.
        my_executor new_exec;
        new_exec.is_possibly_blocking = false;
        return new_exec;
      }
    };

The polymorphic executor has this kind of runtime presence for nearly all of its properties, except for the interface changing ones.

If we ask the question "is P present in E?" at compile time, there are three possible answers:

* Yes.
* Maybe, presence cannot be determined until runtime.
* No.

Obviously, if we ask the question at runtime, there are only two possible answers: Yes or No.

Our current formulation of `is_supportable<>` asks the "Is P present in E?" question at compile time, but does not let us distinguish all three possible answers:
* `P::is_supportable<E> == false` -> Is P present in E? No.
* `P::is_supportable<E> == true` -> Is P present in E? Yes or Maybe.

Using `can_query_v<E, P>` suffers from a similar limitation:
* `can_query_v<E, P> == false` -> Is P present in E? No.
* `can_query_b<E, P> == true` -> Is P present in E? No (if additional syntactic requirements apply), Yes or Maybe.

The constexpr-detection of `P.is_present(E)` does let us distinguish all three:

* `P.is_present(E)` is a constant expression and false -> Is P present in E? No.
* `P.is_present(E)` is a constant expression and true -> Is P present in E? Yes.
* `P.is_present(E)` is not a constant expression -> Is P present in E? Maybe.

However, there may be other reasons for not wanting to (ab)use constexpr like this.

What we need is a compile time way to distinguish all three cases at compile time. Maybe we can do this by replacing `is_supportable` with a type alias `is_present_at_compile_time` and making it conditionally contain a `value` member. E.g.:

    #include <experimental/execution>
    #include <iostream>

    namespace execution = std::experimental::execution;
    using execution::can_query;
    using execution::can_query_v;

    struct compile_time_property_t
    {
      template<class Executor>
        using is_present_at_compile_time =
          can_query<Executor, compile_time_property_t>;
    };

    struct runtime_property_t
    {
      struct maybe {};

      template<class Executor>
        using is_present_at_compile_time =
          std::conditional_t<
            !can_query_v<Executor, runtime_property_t>,
              std::false_type, maybe>;
    };

    struct my_executor1
    {
      bool runtime_value = true;
      int query(compile_time_property_t) const { return 0; }
      bool query(runtime_property_t) const { return runtime_value; }
    };

    struct my_executor2
    {
    };

    template<class Executor, class Property>
    auto is_present_tester(const Executor& e, const Property& p,
        std::enable_if_t<Property::template is_present_at_compile_time<Executor>::value>* = 0)
    {
      std::cout << "Compile time presence? Yes\n";
    }

    template<class Executor, class Property>
    auto is_present_tester(const Executor& e, const Property& p,
        std::enable_if_t<!Property::template is_present_at_compile_time<Executor>::value>* = 0)
    {
      std::cout << "Compile time presence? No\n";
    }

    template<class Executor, class Property>
    void is_present_tester(const Executor& e, const Property& p, ...)
    {
      std::cout << "Compile time presence? Maybe\n";
    }

    int main()
    {
      my_executor1 ex1;
      is_present_tester(ex1, compile_time_property_t(), 0);
      is_present_tester(ex1, runtime_property_t(), 0);
      my_executor2 ex2;
      is_present_tester(ex2, compile_time_property_t(), 0);
      is_present_tester(ex2, runtime_property_t(), 0);
    }

jaredhoberock:
Posted some more thoughts regarding my previous comment at #334.

chriskohlhoff:
I believe your suggested static_query can do the same tri-state distinction. I.e. we can read my "present" as being the same as "the property has some well-known value" which may or may not be testable at compile time.

However, I think my definition of "present" fits better with how we also talk about properties being mutually exclusive. So I suggest we restrict our use of the term "present" to those places only. E.g. this:

> The `possibly_blocking_t`, `always_blocking_t` and `never_blocking_t` properties are mutually exclusive.

becomes something like:

The possibly_blocking_t`, `always_blocking_t` and `never_blocking_t` properties are considered present if their corresponding queries return `true`, and their presence is mutually exclusive.

chriskohlhoff:
Thanks. May want to merge the additional change in #335 too.

PR #328, opened 2018-02-07
Simplified general executor requirements based on LEWG feedback

### LINKED ISSUE #298: Consider simplifying the general executor requirements
There was a suggesting to replace the general executor requirements with references to MoveConstructible, CopyConstructible, & EqualityComparable, with an explanation about what EqualityComparable means for executors.

jaredhoberock:
Fixes #298


jaredhoberock:
Here's what the general requirements look like now:

> An executor type shall satisfy the requirements of `CopyConstructible` (C++Std [copyconstructible]), `Destructible` (C++Std [destructible]), and `EqualityComparable` (C++Std [equalitycomparable]).
> 
> None of these concepts' operations, nor an executor type's swap operations, shall exit via an exception.
> 
> None of these concepts' operations, nor an executor type's associated execution functions, associated query functions, or other member functions defined in executor type requirements, shall introduce data races as a result of concurrent calls to those functions from different threads.
> 
> An executor type's destructor shall not block pending completion of the submitted function objects. [*Note:* The ability to wait for completion of submitted function objects may be provided by the associated execution context. *--end note*]

Note that the operations mentioned in paragraph 3 is not a superset of those from paragraph 2 because `swap` is missing. Is that intentional, or do we want to ban data races in `swap`?

PR #329, opened 2018-02-07
Update allocator properties to use new property form.

AerialMantis:
* Update allocator properties to use new property form.
* Replace default_allocator_t with allocator_t<void>.
* Add note that query should accept both allocator_t<ProtoAllocator> and
allocator_t<void>.
* Update requirements on the static_thread_pool executor to support new
property definitions.


AerialMantis:
@jaredhoberock @chriskohlhoff I've updated this PR with the changes we discussed.

wording.md
```diff
@@ -627,23 +623,43 @@ agents. *--end note*]
...
+        template <typename OtherProtoAllocator>
+        allocator_t<OtherProtoAllocator> operator()(const OtherProtoAllocator &a) const {
+        	return allocator_t<OtherProtoAllocator>{a};
+        }
 
-[*Note:* It is permitted for an allocator provided via the `allocator(ProtoAllocator)` property to be the same type as the allocator provided by the `default_allocator` property. *--end note*]
+        static constexpr bool value() const { return a; }
+    };
+
+| Property | Requirements | is_supportable                      |
```
chriskohlhoff: I think the `is_supportable` column can be deleted

wording.md
```diff
@@ -1555,6 +1575,17 @@ that is defined as a mutually exclusive set, any other properties in the group
...
 
+```
+see-below require(const execution::allocator_t<void>& a) const;
+```
+
+*Returns:* An executor object of an unspecified type conforming to these
+specifications, associated with the same thread pool as `*this`, with the
+`execution::allocator_t<void>` property established such that
+allocation and deallocation associated with function submission will be
+performed using a copy of an implementation deined default allocator. All other
```
chriskohlhoff: Typo: "deined" -> "defined."

I also came up with a cheeky short way of specifying this function in terms of the other allocator_t `require` overload:

*Returns:* `require(execution::allocator(x))`, where `x` is an implementation-defined default allocator.

wording.md
AerialMantis: Done.

wording.md
AerialMantis: Fixed the typo.

Yeah that's good, I've changed it to use that wording.

AerialMantis:
@chriskohlhoff thanks for the feedback, @jaredhoberock I think this can be merged now if you're happy with it.

jaredhoberock:
Looks good, merged.

Issue #276, opened 2017-11-07
static_thread_pool executors need .query()


jaredhoberock:
Seems to have been an oversight in R3.

AerialMantis:
I thought we had added that, yeah must have been an oversight, I can create a pull request for this.

Do we want to require that the `static_thread_pool` executor types have a `query` overload for each property which it has a `require` overload, except the interface altering properties?

```
query_member_result_t<C, execution::caller_parallel_execution_t>
  query(execution::caller_parallel_execution_t) const;
query_member_result_t<C, execution::bulk_parallel_execution_t>
  query(execution::bulk_parallel_execution_t) const;
query_member_result_t<C, execution::thread_execution_mapping_t>
  query(execution::thread_execution_mapping_t) const;
query_member_result_t<C, execution::never_blocking_t>
  query(execution::never_blocking_t) const;
query_member_result_t<C, execution::possibly_blocking_t>
  query(execution::possibly_blocking_t) const;
query_member_result_t<C, execution::always_blocking_t>
  query(execution::always_blocking_t) const;
query_member_result_t<C, execution::continuation_t>
  query(execution::continuation_t) const;
query_member_result_t<C, execution::not_continuation_t>
  query(execution::not_continuation_t) const;
query_member_result_t<C, execution::outstanding_work_t>
  query(execution::outstanding_work_t) const;
query_member_result_t<C, execution::not_outstanding_work_t>
  query(execution::not_outstanding_work_t) const;
template<class ProtoAllocator>
  query_member_result_t<C, execution::allocator_wrapper_t>
  query(const execution::allocator_wrapper_t<ProtoAllocator>& a) const;
```

jaredhoberock:
As a general rule, I think standard "good citizen" executors ought to support the customization points to the fullest degree possible. An exception to this rule would be cases where support could be difficult to implement efficiently or safely.

PR #330, opened 2018-02-07
Issue #276: Add requirement for query members on static_thread_pool.

AerialMantis:
* Add requirement on static_thread_pool to provide query member function
overloads for all standard properties.

### DIFF
--- a/wording.md
+++ b/wording.md
@@ -1643,13 +1643,6 @@ class C
 
     // executor operations:
 
-    C require(execution::oneway_t) const;
-    C require(execution::twoway_t) const;
-    C require(execution::then_t) const;
-    C require(execution::single_t) const;
-    C require(execution::bulk_t) const;
-    C require(execution::bulk_parallel_execution_t) const;
-    C require(execution::thread_execution_mapping_t) const;
     see-below require(execution::never_blocking_t) const;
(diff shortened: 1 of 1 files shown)


chriskohlhoff:
As discussed on the call, the following query members are not required:

    bool query(execution::oneway_t) const;
    bool query(execution::twoway_t) const;
    bool query(execution::then_t) const;
    bool query(execution::single_t) const;
    bool query(execution::bulk_t) const;

as these will be determined automatically by the `static_query_v` test for those properties. While you're there you might consider removing the corresponding `require` members, as they will never be used (due to the new `static_query_v` bullet point in the `require` customisation point).

I also suggest making the queries for `thread_execution_mapping_t` and `bulk_parallel_execution_t` into static constexpr functions, e.g.:

    static constexpr bool query(execution::thread_execution_mapping_t);

with specification that it always return true. This will also enable `static_query_v` for this property and consequently make `require` an identity operation for it.

AerialMantis:
@chriskohlhoff thanks for the suggestion, makes sense, I've included this change as well.
@jaredhoberock I think this is ready to merge if you're happy with it.

jaredhoberock:
Thanks, merged.

Issue #317, opened 2018-01-11
Proposed wording for executor equality


AerialMantis:
After discussing what it means for an executor to be equal I started thinking about some of the issues we discussed and thought I'd have a go at what the potential wording might look like.

**Non-polymorphic executors**

Now that the context is simply another optional property to an executor it seems like equality comes down to two factors.

For any two executors X and Y:
* Is the type of X the same as the type of Y?
* Does X contain the same set of properties as Y?

So from these questions, we can derive the following rules:
1. If the type of X is not the same as the type of Y then X and Y cannot be equal because even if X and Y were to have the same set of properties, they could be implemented differently or be provided through adaptations.
2. If X does not contain the same set of properties as Y then X and Y cannot be equal because the difference in properties could enforce different semantics.
3. If the type of X is the same as the type of Y and X contains the same set of properties as Y then you can be sure that X and Y are equal as they have the same properties which means they will enforce the same semantics and they are the same type which means they will be implemented in the same way.

From these rules we can derive the definition:

> For any two non-polymorphic executors; X and Y, X is equal to Y if the type of X is the same as the type of Y and X contains the same set of properties as Y.

**Polymorphic executors**

Now if we look at equality between polymorphic executors, this becomes more complicated, as we also have to consider the type of polymorphic executors as well as the executors which they wrap.

We can assume that the previous rules 1 and 2 still apply for the executors wrapped within the polymorphic executors, however, how does rule 3 still apply in the context of polymorphic executors? We could potentially allow two polymorphic executors of different types to still be equally providing the wrapped executors are the same type and contain the same set of properties, however, this would require constraints on the polymorphic executor such that it cannot adapt the semantics of the wrapped executor in any way and would also require that ordering of properties have no bearing on the semantics which is still an ongoing discussion. Therefore, to keep things simple for now we can require the polymorphic executors to be the same type.

For any two polymorphic executors PX and PY wrapping executors X and Y respectively:
* Is the type of PX the same as the type of PY?
* Is the type of X the same as the type of Y?
* Does X contain the same set of properties as Y?

So from these questions, we can derive the following rules:
1. If the type of X is not the same as the type of Y then PX and PY cannot be equal because even if X and Y were to have the same set of properties, they could be implemented differently or be provided through adaptations.
2. If X does not contain the same set of properties as Y then PX and PY cannot be equal because the difference in properties could enforce different semantics.
[...]

PR #331, opened 2018-02-08
Issue #317: Add equality requirements based on properties.

AerialMantis:
* Extend the equality requirements for executors to include a
requirement on the result of query for each property.


AerialMantis:
@chriskohlhoff @jaredhoberock as part of the merge https://github.com/executors/executors/pull/328 I moved the equality requirements to a separate paragraph, added the additional requirement that the property must be equality comparable and reversed the conditional in the statement as we discussed. Should I also put a note in here about properties of an executor not exposed through a query?

wording.md
```diff
@@ -187,6 +187,8 @@ None of these concepts' operations, nor an executor type's swap operations, shal
 
 None of these concepts' operations, nor an executor type's associated execution functions, associated query functions, or other member functions defined in executor type requirements, shall introduce data races as a result of concurrent calls to those functions from different threads.
 
+For any two (possibly const) values `x1` and `x2` of some executor type `X`, if `x1 == x2` returns `true` then `x1.query(p)` and `x2.query(p)` must both be well-formed and result in a non-void type that is `EqualityComparable` (C++Std [equalitycomparable]) and `x1.query(p) == x2.query(p)`, for each property `p` where either `x1.query(p)` or `x2.query(p)` is well-formed. [*Note:* The above requirements imply that `x1 == x2` returns `true` if `x1` and `x2` can be interchanged with identical effects. Returning `false` does not necessarily imply that the effects are not identical. *--end note*]
```
chriskohlhoff: I suggest reordering the clauses in the sentence. How about:

For any two (possibly const) values `x1` and `x2` of some executor type `X`, `x1 == x2` shall return `true` only if `x1.query(p) == x2.query(p)` for every property `p` where both `x1.query(p)` or `x2.query(p)` are well-formed and result in a non-void type that is `EqualityComparable` (C++Std [equalitycomparable]). [*Note:* The above requirements imply that `x1 == x2` returns `true` if `x1` and `x2` can be interchanged with identical effects. Returning `false` does not necessarily imply that the effects are not identical. *--end note*]

chriskohlhoff:
Yep, the extra note would be good.

wording.md
jaredhoberock: I was about to ask whether or not we want to say something about properties whose values are not `EqualityComparable`, but after thinking about it, I think we don't want to do that. If the properties values aren't comparable, then they shouldn't have any say in executor equality. So, Chris's suggested change seems good to me.

AerialMantis:
I'm thinking about the best way to phrase the note for properties not exposed by query, my current thinking is something like:
```
An executor may contain additional properties which are not exposed by a named property type which can be observed via `execution::query`, however, these properties may not conflict with any behavior associated with a named property type, and therefore do not affect equality.
```
Is this what we want or is this too restrictive, do we want to allow an executor to have non-exposed properties which can affect equality, but are just not observable?

chriskohlhoff:
It might help to have an extra word near the start like "An executor may logically contain additional properties" or "An executor may conceptually contain additional properties".

We don't want to restrict in this case, so how about saying something like it is up to the concrete executor implementation to determine whether these properties affect equality.

wording.md

AerialMantis:
I agree, I've added the changes you suggested.

wording.md
```diff
@@ -187,6 +187,10 @@ None of these concepts' operations, nor an executor type's swap operations, shal
...
 
+For any two (possibly const) values `x1` and `x2` of some executor type `X`, `x1 == x2` shall return `true` only if `x1.query(p) == x2.query(p)` for every property `p` where both `x1.query(p)` and `x2.query(p)` are well-formed and result in a non-void type that is `EqualityComparable` (C++Std [equalitycomparable]). [*Note:* The above requirements imply that `x1 == x2` returns `true` if `x1` and `x2` can be interchanged with identical effects. An executor may conceptually contain additional properties which are not exposed by a named property type that can be observed via `execution::query`, in this case, it is up to the concrete executor implementation to decide if these properties affect equality. Returning `false` does not necessarily imply that the effects are not identical. *--end note*]
+
+these properties may not conflict with any behavior associated with a named property type, and therefore do not affect equality.
```
chriskohlhoff: Was there supposed to be an extra line before this?

AerialMantis: This was left in by mistake, I've removed it.

wording.md
```diff
@@ -187,6 +187,8 @@ None of these concepts' operations, nor an executor type's swap operations, shal
 
 None of these concepts' operations, nor an executor type's associated execution functions, associated query functions, or other member functions defined in executor type requirements, shall introduce data races as a result of concurrent calls to those functions from different threads.
 
+For any two (possibly const) values `x1` and `x2` of some executor type `X`, `x1 == x2` shall return `true` only if `x1.query(p) == x2.query(p)` for every property `p` where both `x1.query(p)` and `x2.query(p)` are well-formed and result in a non-void type that is `EqualityComparable` (C++Std [equalitycomparable]). [*Note:* The above requirements imply that `x1 == x2` returns `true` if `x1` and `x2` can be interchanged with identical effects. An executor may conceptually contain additional properties which are not exposed by a named property type that can be observed via `execution::query`, in this case, it is up to the concrete executor implementation to decide if these properties affect equality. Returning `false` does not necessarily imply that the effects are not identical. *--end note*]
+
```
chriskohlhoff: Tiny punctuation change in 2nd last sentence to use semicolon rather than comma:

>`execution::query`, in this case, ...

 to

> `execution::query`; in this case, ...

AerialMantis:
@jaredhoberock I think this is ready to merge if you're happy with it.

jaredhoberock:
Looks good. Merged.

PR #332, opened 2018-02-08
Add requirement to query that the value cannot change on its own

AerialMantis:
* Add requirement that subsequent calls to query must always return the
same value unless the value has been modified by a call to require or
prefer.

### DIFF
--- a/wording.md
+++ b/wording.md
@@ -431,7 +431,7 @@ The `context_t` property can be used only with `query`, which returns the **exec
 
 An execution context is a program object that represents a specific collection of execution resources and the **execution agents** that exist within those resources. Execution agents are units of execution, and a 1-to-1 mapping exists between an execution agent and an invocation of a callable function object submitted via the executor.
 
-The value returned from `execution::query(e, context_t)`, where `e` is an executor, must not change between calls.
+The value returned from `execution::query(e, context_t)`, where `e` is an executor, shall not change between calls.
 
 ### Interface-changing properties
 
@@ -571,7 +571,7 @@ The `possibly_blocking_t`, `always_blocking_t` and `never_blocking_t` properties
 
 [*Note:* The guarantees of `possibly_blocking_t`, `always_blocking_t` and `never_blocking_t` implies the relationships: `possibly_blocking < always_blocking` and `possibly_blocking < never_blocking` *--end note*]
 
-The value returned from `execution::query(e, p)` must not change between calls unless `e` is assigned another executor which has a different value for `p`. The value returned from `execution::query(e, p1)` and a subsequent call `execution::query(e1, p1)` where:
+The value returned from `execution::query(e, p)` shall not change between calls unless `e` is assigned another executor which has a different value for `p`. The value returned from `execution::query(e, p1)` and a subsequent call `execution::query(e1, p1)` where:
 * `p1` is `execution::possibly_blocking`, `execution::always_blocking` or `execution::never_blocking`, and
 * `e1` is the result of `execution::require(e, p2)` or `execution::prefer(e, p2)`,
 shall compare equal unless:
@@ -634,7 +634,7 @@ The `not_adaptable_blocking_t` and `adaptable_blocking_t` properties are mutuall
 
 [*Note:* The `two_way_t` property is included here as the `require` customization point's `two_way_t` adaptation is specified in terms of `std::experimental::future`, and that template supports blocking wait operations. *--end note*]
 
-The value returned from `execution::query(e, p)` must not change between calls unless `e` is assigned another executor which has a different value for `p`. The value returned from `execution::query(e, p1)` and a subsequent call `execution::query(e1, p1)` where:
+The value returned from `execution::query(e, p)` shall not change between calls unless `e` is assigned another executor which has a different value for `p`. The value returned from `execution::query(e, p1)` and a subsequent call `execution::query(e1, p1)` where:
 * `p1` is `execution::adaptable_blocking` or `execution::not_adaptable_blocking`, and
 * `e1` is the result of `execution::require(e, p2)` or `execution::prefer(e, p2)`,
 shall compare equal unless:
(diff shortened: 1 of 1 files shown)


jaredhoberock:
I think we decided on the last call that this requirement should instead be a per-property requirement common to each of our standard properties, right?

wording.md
```diff
@@ -389,7 +389,7 @@ The current value of an executor's properties can be queried by calling the `que
 
 | Expression | Comments |
 |------------|----------|
-| `x.query(p)` | Returns the current value of the requested property `p`. The value returned from `x.query(p)` must not change between calls, unless the property has been modified by a call to `require` or `prefer`. The expression is ill formed if an executor is unable to return the requested property. |
+| `x.query(p)` | Returns the current value of the requested property `p`. The expression is ill formed if an executor is unable to return the requested property. |
```
jaredhoberock: Small nit: "ill formed" -> "ill-formed"

AerialMantis:
@jaredhoberock I believe so yes, I've moved this requirement to the general properties requirements, and also added a note about which properties can change a value (@chriskohlhoff ).

jaredhoberock:
I believe the way it's currently worded, as part of the general requirements for properties, makes it a requirement for all properties, standard or not. So I think it may need to be moved downward to make it a requirement only for the concrete properties P0443 specifies.

wording.md
```diff
@@ -433,6 +431,8 @@ The `context_t` property can be used only with `query`, which returns the **exec
 
 An execution context is a program object that represents a specific collection of execution resources and the **execution agents** that exist within those resources. Execution agents are units of execution, and a 1-to-1 mapping exists between an execution agent and an invocation of a callable function object submitted via the executor.
 
+The value returned from `execution::query(e, context_t)`, where `e` is an executor, must not change between calls.
```
jaredhoberock: Small nit: "must" -> "shall"

AerialMantis:
Yeah that makes sense. I've moved the property to be per standard property. This means stating the requirements per property section, though it does mean we can be more explicit about which properties can modify a property value. How does this look?

wording.md
```diff
@@ -467,6 +467,8 @@ The directionality properties conform to the following specification:
 
 The `oneway_t`, `twoway_t` and `then_t` properties are not mutually exclusive.
 
+The value returned from `execution::query(e, p)` must not change between calls, unless that value has been modified by a call to `execution::require(e, p)` or `execution::prefer(e, p)`, where `e` is an executor and `p` is `execution::oneway`, `execution::twoway` or `execution::then`.
```
jaredhoberock: I don't think there's any reason to include this paragraph for interface-changing properties because there is no way for an executor to affect the result of e.g. `execution::query(ex, execution::oneway)`.

wording.md
```diff
@@ -513,6 +515,8 @@ The cardinality properties conform to the following specification:
 
 The `single_t` and `bulk_t` properties are not mutually exclusive.
 
+The value returned from `execution::query(e, p)` must not change between calls, unless that value has been modified by a call to `execution::require(e, p)` or `execution::prefer(e, p)`, where `e` is an executor and `p` is `execution::single` or `execution::bulk`.
+
```
jaredhoberock: Same with this paragraph.

chriskohlhoff:
>The value returned from `execution::query(e, p)` must not change between calls, unless that value has been modified by a call to `execution::require(e, p)` or `execution::prefer(e, p)`, ...

It just occurred to me that `execution::require(e, p)` and `execution::prefer(e, p)` don't change the state of `e`. So maybe it should be:

> The value returned from `execution::query(e, p)` shall not change between calls unless modified by assignment of another executor into `e`.

Thoughts?

wording.md
```diff
@@ -571,6 +575,12 @@ The `possibly_blocking_t`, `always_blocking_t` and `never_blocking_t` properties
 
 [*Note:* The guarantees of `possibly_blocking_t`, `always_blocking_t` and `never_blocking_t` implies the relationships: `possibly_blocking < always_blocking` and `possibly_blocking < never_blocking` *--end note*]
 
+The value returned from `execution::query(e, p1)` must not change between calls, unless that value has been modified by a call to `execution::require(e, p2)` or `execution::prefer(e, p2)`, where:
+* `e` is an executor,
+* `p1` is `execution::possibly_blocking`, `execution::always_blocking` or `execution::never_blocking`,
+* `p2` is `execution::possibly_blocking`, `execution::always_blocking` or `execution::never_blocking`,
+* and `p1` and `p2` are different types.
+
```
jaredhoberock: And this paragraph.

jaredhoberock:
Yes, I think that's right.

chriskohlhoff:
Actually that's not quite it either. We want to say something about the result of require and prefer and how it relates to the source. So we may also need something like:

>The value returned from `execution::query(e, p)` and a subsequent call to `execution::query(e1, p)` shall compare equal, where `e1` was produced by a call to `execution::require(e, p)` or `execution::prefer(e, p)`, unless `p` is...

AerialMantis:
So my thinking was this wording would only apply in the case where the same executor type is returned from `require` or `prefer`, since if they return a different executor type then the value isn't really changing. I agree that in that case the wording for the interface changing properties we don't need this as they always return a new executor type unless it already has the property.

AerialMantis:
I see what you mean, I'll reword this to incorporate the return value of `require` or `prefer`.

chriskohlhoff:
I think it needs both as separate statements. i.e. 1) only assignment can modify; 2) preservation of property across require/prefer

AerialMantis:
How does this sound?

The value returned from `execution::query(e, p)` must not change between calls unless `e` is assigned another executor which has a different value for `p`. The value returned from `execution::query(e, p1)` and a subsequent call `execution::query(e1, p1)` where `e1` is the result of `execution::require(e, p2)` or `execution::prefer(e, p2)` shall compare equal unless:
* `p1` is `execution::possibly_blocking`, `execution::always_blocking` or `execution::never_blocking`,
* `p2` is `execution::possibly_blocking`, `execution::always_blocking` or `execution::never_blocking`,
* and `p1` and `p2` are different types.

chriskohlhoff:
Almost. I think we want to constrain p1 to be one of the three properties, rather than include it in the conditions for equality, so:

The value returned from `execution::query(e, p)` must not change between calls unless `e` is assigned another executor which has a different value for `p`. The value returned from `execution::query(e, p1)` and a subsequent call `execution::query(e1, p1)`, where:
* `p1` is `execution::possibly_blocking`, `execution::always_blocking` or `execution::never_blocking`, and
* `e1` is the result of `execution::require(e, p2)` or `execution::prefer(e, p2)`
shall compare equal unless
* `p2` is `execution::possibly_blocking`, `execution::always_blocking` or `execution::never_blocking`, and

chriskohlhoff:
Also s/must/shall.

AerialMantis:
Yeah, that makes sense, I've added that new wording with the change you suggested.

wording.md

jaredhoberock:
Anything else to do on this pull request?

AerialMantis:
No, if everyone is happy with this I think it's ready to merge.

PR #333, opened 2018-02-08
WIP: Refactoring the polymorphic executor

AerialMantis:
**Changes**

* Make polymorphic executor templated on a variadic parameter pack of
property types.
* Introduce generic require, prefer and query member function templates
for the polymorphic executor.
* Add requirements on polymorphic executor require, prefer and query which require properties to be requireable, preferable and queryable according to new property form.
* Add requirements on polymorphic executor constructor which require
properties to be requireable, preferable and queryable according to new
property form.
* Remove requirement that execution functions call require() before
delegating to the target executor.
* Add missing template parameters from non-member functions.
* Remove prefer member function.
* Change e.require(p) to execution::require(e, p)
* Change e.query(p) to execution::query(e, p)
* Chnage Properties to SupportableProperties.
* Use correct wording for describing when a function should not
participate in overload resolution.
* Add CONTAINS_PROPERTY macro for checking when a property type exists
within a template parameter pack.
* Fix the wording for the return value of require and prefer.
* Add requirement that the interface changing properties be required on
construction.
* Add narrowing conversion constructor.
Add note to start of polymorphic executor wrapper section stating that
it had been written in a way which can be separated if necessary.

### DIFF
--- a/wording.md
+++ b/wording.md
@@ -149,6 +149,8 @@ namespace execution {
   // Polymorphic executor wrappers:
 
   class bad_executor;
+
+  template <class... SupportableProperties>
   class executor;
 
 } // namespace execution
@@ -801,6 +803,8 @@ This sub-clause contains templates that may be used to query the properties of a
 
 ## Polymorphic executor wrappers
 
+The polymorphic executor wrapper class has been written such that it could be separated from the rest of the proposal if nessesary.
+
 ### Class `bad_executor`
 
 An exception of type `bad_executor` is thrown by `executor` member functions `execute`, `twoway_execute`, `bulk_execute`, and `bulk_twoway_execute` when the executor object has no target.
@@ -829,7 +833,10 @@ bad_executor() noexcept;
 The `executor` class template provides a polymorphic wrapper for executor types.
 
 ```
-template <class... Properties>
+// macro for testing whether a property parameter pack contains a particular property
+#define CONTAINS_PROPERTY(property, property_parameter_pack) td::disjunction_v<std::is_same<property, property_parameter_pack>...>
+
+template <class... SupportableProperties>
 class executor
 {
 public:
@@ -842,6 +849,7 @@ public:
   executor(const executor& e) noexcept;
   executor(executor&& e) noexcept;
   template<class Executor> executor(Executor e);
+  template<class... OtherSupportableProperties> executor(executor<OtherSupportableProperties...> e);
 
   executor& operator=(const executor& e) noexcept;
   executor& operator=(executor&& e) noexcept;
@@ -858,9 +866,6 @@ public:
 
   template <class Property>
   executor require(Property) const;
-
-  template <class Property>
-  executor prefer(Property) const;
 
   template <class Property>
   executor query(Property) const;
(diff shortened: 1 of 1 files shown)


AerialMantis:
@jaredhoberock @chriskohlhoff  I've finished the polymorphic executor refactor and updated it with the changes we discussed.

wording.md
```diff
@@ -801,6 +803,8 @@ This sub-clause contains templates that may be used to query the properties of a
 
 ## Polymorphic executor wrappers
 
+The polymorphic executor wrapper class has been written such that it could be separated from the rest of the proposal if nessesary.
```
chriskohlhoff: Typo in "necessary".

Also suggest wrapping this in

    [*Commentary: ... text here ... *]

wording.md
```diff
@@ -824,11 +828,15 @@ bad_executor() noexcept;
...
-### Class `executor`
+### Class template `executor`
 
-The `executor` class provides a polymorphic wrapper for executor types.
+The `executor` class template provides a polymorphic wrapper for executor types.
 
 ```
+// macro for testing whether a property parameter pack contains a particular property
+#define CONTAINS_PROPERTY(property, property_parameter_pack) td::disjunction_v<std::is_same<property, property_parameter_pack>...>
+
```
chriskohlhoff: Sorry for the ambiguity. I didn't mean an actual preprocessor macro, but a "standardese" one :)

I think something like the wording from [thread.decaycopy] is needed a couple of lines above. So something like:

> In several places in this section the operation CONTAINS_PROPERTY is used. All such uses mean `std::disjunction_v<std::is_same<property, property_parameter_pack>...>`.

wording.md
```diff
@@ -1026,40 +1032,37 @@ void swap(executor& other) noexcept;
...
-const context_type& query(context_t) const noexcept;
+template <class Property>
+executor require(Property p) const;
 ```
 
-*Requires:* `*this != nullptr`.
+*Remarks:* This function shall not participate in overload resolution unless: `CONTAINS_PROPERTY(Property, SupportableProperties) && Property::is_requirable`.
 
-*Returns:* A polymorphic execution context wrapper whose target is `e.query(execution::context)`, where `e` is the target object of `*this`.
+*Returns:* A polymorphic wrapper whose target is is the result of `execution::require(e, p)`, where `e` is the target object of `*this`.
```
chriskohlhoff: double "is is" here

wording.md
```diff
@@ -1158,45 +1170,47 @@ bool operator==(const executor& a, const executor& b) noexcept;
...
-executor prefer(const executor& e, bulk_unsequenced_execution_t p);
-executor prefer(const executor& e, new_thread_execution_mapping_t p);
+template <class Property, class... SupportableProperties>
+executor prefer(const executor<SupportableProperties...>& e, Property p);
 ```
 
-*Returns:* A polymorphic wrapper whose target is `execution::prefer(e1, p)`, where `e1` is the target object of `e`.
+*Remarks:* This function shall not participate in overload resolution unless: `CONTAINS_PROPERTY(Property, SupportableProperties)`.
+
+*Returns:* A polymorphic wrapper whose target is is the result of `execution::prefer(e, p)`, where `e` is the target object of `*this`.
```
chriskohlhoff: another "is is"

chriskohlhoff:
Looks good to me overall, just those tiny issues i commented on.

wording.md

wording.md
AerialMantis: I had a feeling I this wasn't right, thanks :)

I also changed `property` and `property_parameter_pack` to `p` and `pn`  to make it more concise.

wording.md

wording.md

jaredhoberock:
Thanks Gordon, is this ready to merge?

AerialMantis:
@chriskohlhoff thanks for the feedback.
@jaredhoberock yeah I just pushed the fixes so I think this is ready to go if you're happy with it.

Issue #334, opened 2018-02-08
Distinguishing static property queries from dynamic property queries


jaredhoberock:
Here's an alternate properties formulation which attempts to address [my previous comment on ChrisK's pull request](https://github.com/executors/executors/pull/325#issuecomment-364222431). It's essentially a hybrid between what ChrisK has done in his PR and my prototype implementation of P0443.

#### Definitions:

1. An executor type `E` has static property `P` if it is possible to query `E` for `P` to retrieve an associated value.

2. An executor instance `e` has property `P` if it is possible to query `e` for `P` to retrieve an associated value.

3. `e` or `E` has a property `P` with a value `v` if `e` or `E`'s value associated with `P` is equivalent to `v`.

#### Queries

1. To query `E` for a static property `P`:

       template<class E, class P>
       constexpr auto static_property_value_v = P::template static_query_v<E>;

2. Let `p` be an instance of `P`. To query `e` for property `P`:

       p(e);

#### Property detection

An executor `E` has a static property `P` if `static_property_value_v<E,P>` is well-formed:

       template<class E, class P>
       bool has_static_property_v = is_detected_v<static_property_value_v, E, P>;

An executor e has a property P if either `has_static_property_v<E,P>` or `std::declval<P>()(std::declval<const E&>())` is well-formed:

       template<class E, class P>
       bool has_property_v = has_static_property_v<E,P> or is_detected_v<__call_operator_t, P, const E&>;

#### Property value comparisons

1. To compare `E`'s value associated with static property `P` with a value `v`:

       static_property_value_v<E,P> == v;

2. Let `p` be an instance of `P`. To compare `e`'s value associated with property `P` with a value `v`:

       p(e) == v;

#### Postconditions for execution::require(e, p)

1. `has_property_v<decltype(result), decltype(p)>` is true

2. if `p(e) == p.value()` is well-formed, `p(e) == p.value()` is true
  
    
----

The good:
1. Seems to account for every kind of query we care about
2. Can enforce `execution::require()`'s post-conditions
3. Executor types can't "hijack" interface-changing properties via a free function
4. Common-case query syntax like `execution::context(e)` feels more ergonomic than `execution::query(e, execution::context)`
5. Doesn't require the bespoke `is_oneway_executor` et al traits
6. Doesn't require `execution::query()`

The bad:
1. In order to test the value associated with an executor's property with a particular value, the type of that value must be `EqualityComparable`
[...]

PR #335, opened 2018-02-09
Property static query

chriskohlhoff:
Alternative to #325, addresses #334.


chriskohlhoff:
I think this one's ready to merge too. It should apply just the additional static_query_v change.

PR #339, opened 2018-02-11
Add prefer_only property adapter.

chriskohlhoff:


jaredhoberock:
I think some of the members of `prefer_only` are conditionally present depending on whether or not the members are present in `InnerProperty`. These would be `polymorphic_query_result_type`, `static_query_v`,  and `value()`. Do we need to specify that the types are conditionally present and that `.value()` does not participate in overload resolution if `property.value()` is ill-formed?

I see the following compile errors when `prefer_only` is instantiated with a minimal property type:

```
template<class InnerProperty>
struct prefer_only
{
  InnerProperty property;

  static constexpr bool is_preferable = InnerProperty::is_preferable;

  using polymorphic_query_result_type =
    typename InnerProperty::polymorphic_query_result_type;

  template<class Executor>
    static constexpr auto static_query_v =
      InnerProperty::template static_query_v<Executor>;

  constexpr prefer_only(const InnerProperty& p);

  constexpr auto value() const
    noexcept(noexcept(property.value()))
      -> decltype(property.value());

//  template<class Executor>
//  friend auto prefer(Executor ex, const prefer_only& p)
//    noexcept(noexcept(execution::prefer(std::move(ex), std::declval<const InnerProperty>())))
//      -> decltype(execution::prefer(std::move(ex), std::declval<const InnerProperty>()));
//
//  template<class Executor>
//  friend constexpr auto query(const Executor& ex, const prefer_only& p)
//    noexcept(noexcept(execution::query(ex, std::declval<const InnerProperty>())))
//      -> decltype(execution::query(ex, std::declval<const InnerProperty>()));
};

struct my_property
{
  static constexpr bool is_preferable = true;
};

int main()
{
  prefer_only<my_property> p{my_property()};
}
```

```
$ clang -std=c++14 test.cpp 
test.cpp:10:29: error: no type named 'polymorphic_query_result_type' in 'my_property'
    typename InnerProperty::polymorphic_query_result_type;
    ~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
test.cpp:41:28: note: in instantiation of template class 'prefer_only<my_property>' requested here
  prefer_only<my_property> p{my_property()};
                           ^
test.cpp:14:31: error: 'static_query_v' following the 'template' keyword does not refer to a template
      InnerProperty::template static_query_v<Executor>;
      ~~~~~~~~~~~~~~~         ^
test.cpp:20:28: error: no member named 'value' in 'my_property'
      -> decltype(property.value());
                  ~~~~~~~~ ^
3 errors generated.


```

chriskohlhoff:
Yes, I'll fix that now.

jaredhoberock:
Thanks, looks good.

PR #341, opened 2018-02-11
Simplify property requirement tables

### LINKED ISSUE #336: Cleanup suggestion
I would submit a pull request for this, but don't want to do so before the larger reorganization change is complete.

I was looking for places where `is_oneway_executor` et al. are used in the paper (in case they can be safely eliminated), and they only seem to show up in the tables summarizing property requirements. For example,

| Property | Requirements | static_query_v value           |
|----------|--------------| -------------------------------|
| `oneway_t` | The executor type satisfies the `OneWayExecutor` or `BulkOneWayExecutor` requirements. | `is_oneway_executor_v<Executor> || is_bulk_oneway_executor_v<Executor>` |
| `twoway_t` | The executor type satisfies the `TwoWayExecutor` or `BulkTwoWayExecutor` requirements. | `is_twoway_executor_v<Executor> || is_bulk_twoway_executor_v<Executor>` |
| `then_t` | The executor type satisfies the `ThenExecutor` or `BulkThenExecutor` requirements. | `is_then_executor_v<Executor> || is_bulk_then_executor_v<Executor>` |

In these tables, I think the value of `static_query_v` is true when the Requirements are met, and false otherwise. Are there any exceptions? If not, I think we could probably remove the `static_query_v` value column and make a general statement instead.

jaredhoberock:
Fixes #336

### DIFF
--- a/wording.md
+++ b/wording.md
@@ -456,11 +456,13 @@ The directionality properties conform to the following specification:
       static constexpr bool value() const { return true; }
     };
 
-| Property | Requirements | static_query_v value           |
-|----------|--------------| -------------------------------|
-| `oneway_t` | The executor type satisfies the `OneWayExecutor` or `BulkOneWayExecutor` requirements. | `is_oneway_executor_v<Executor> || is_bulk_oneway_executor_v<Executor>` |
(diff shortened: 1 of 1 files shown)


chriskohlhoff:
Looks good to me. I'll let you merge to avoid conflicts.

PR #357, opened 2018-02-26
Introduce missing adaptations

jaredhoberock:
As discussed on the last call, this is a WIP to add wording which would introduce adaptations that are absent from P0443R5.

The first change introduces an adaptation which can add the `twoway` property to executors with the `then` property.


jaredhoberock:
When reviewing these changes we ought to decide whether we want to specify how the adaptations are implemented. At the moment, the adaptations are specified to be implemented in terms of an extant execution function.

An argument for specifying it is that it makes it easier to understand how the various member functions actually work. An argument against specifying the mechanics of the adaptation would allow the implementation the freedom to choose from various possibilities, as in the case of `require(twoway, ex)`.

jaredhoberock:
> When reviewing these changes we ought to decide whether we want to specify how the adaptations are implemented. At the moment, the adaptations are specified to be implemented in terms of a specific, preexisting execution function.

The consensus that we arrived at during the March 2 meeting was to produce a TS with these adaptations specified. If this creates problems, we can discover them during experimentation with TS implementations.

jaredhoberock:
@chriskohlhoff @AerialMantis Did we say we want to treat this PR as a "bug fix" or is it a substantial enough change that we want LEWG to sign off first?

chriskohlhoff:
I think it is substantive enough to require LEWG blessing (on the basis that if we didn't include it the proposal still works, so it can't technically be considered a bug fix).

jaredhoberock:
@chriskohlhoff @AerialMantis @ccmysen There are no merge conflicts on this PR, so I think the wording is ready for review.

wording.md
```diff
@@ -516,6 +519,26 @@ The cardinality properties conform to the following specification:
...
+
+    struct single_t
+    {
+      template<class Executor>
+        friend see-below require(Executor ex, single_t);
+    };
+
+This customization point returns an executor that satisfies the `single_t` requirements by adapting the native functionality of an executor that does not satisfy the `single_t` requirements.
+
+*Returns:* A value `e1` of type `E1` that holds a copy of `ex`. `E1` has member functions `require` and `query` that forward to the corresponding members of the copy of `ex`, if present. For some type `T`, the type yielded by `executor_future_t<E1, T>` is `executor_future_t<Executor, T>` if `twoway_t::static_query_v<Executor> || then_t::static_query_v<Executor>` is true; otherwise, it is `std::experimental::future<T>`. `e1` has the same properties as `ex`, except for the addition of the `twoway_t` property. Additional `single_t` requirements are satisfied as follows:
```
chriskohlhoff: > except for the addition of the `twoway_t` property. 

Should be `single_t`.

Issue #352, opened 2018-02-16
prefer_only causes infinitely recursive template instantiation


chriskohlhoff:
Due to the interaction between the `prefer_only` constructor:

    constexpr prefer_only(const InnerProperty& p);

which enables implicit conversion from the inner property type, and the specification of the `prefer` and `query` friends:

      template<class Executor>
      friend auto prefer(Executor ex, const prefer_only& p)
        noexcept(noexcept(execution::prefer(std::move(ex), std::declval<const InnerProperty>())))
          -> decltype(execution::prefer(std::move(ex), std::declval<const InnerProperty>()));

      template<class Executor>
      friend constexpr auto query(const Executor& ex, const prefer_only& p)
        noexcept(noexcept(execution::query(ex, std::declval<const InnerProperty>())))
          -> decltype(execution::query(ex, std::declval<const InnerProperty>()));

we can get infinitely recursive template instantiation. I.e. the friend `prefer` calls the customisation point which calls the same friend `prefer` again, etc.

I think we can fix this by specifying the friend functions with a `Property` template parameter:

      template<class Executor, class Property>
      friend auto prefer(Executor ex, const Property& p)
        noexcept(noexcept(execution::prefer(std::move(ex), std::declval<const InnerProperty>())))
          -> decltype(execution::prefer(std::move(ex), std::declval<const InnerProperty>()));

      template<class Executor, class Property>
      friend constexpr auto query(const Executor& ex, const Property& p)
        noexcept(noexcept(execution::query(ex, std::declval<const InnerProperty>())))
          -> decltype(execution::query(ex, std::declval<const InnerProperty>()));

and extending the "*Remarks:* Shall not participate in overload resolution unless..." to include the requirement that `std::is_same<Property, prefer_only>` is `true`.

Issue #347, opened 2018-02-13
Mutually-exclusive properties cleanup idea


jaredhoberock:
In various places in P0443R5, we say that certain sets of properties are mutually-exclusive. I would like to find a way to use the C++ type system to enforce these requirements and remove such statements. If we can do so, we should be able to completely eliminate the possibility of an executor author accidentally implementing `.query(always_blocking_t) == .query(never_blocking_t)` and similar bugs.

My basic idea is to make sets of properties such as `possibly_blocking`, `always_blocking`, and `never_blocking` instances of the same underlying `blocking_t` type with differing values:

```
constexpr blocking_t blocking{};
constexpr blocking_t possibly_blocking{blocking_t::possibly};
constexpr blocking_t always_blocking{blocking_t::always};
constexpr blocking_t never_blocking{blocking_t::never};
```

#### Blocking properties

    class blocking_t
    {
      public:
        enum guarantee {possibly, always, never};
  
        static constexpr bool is_requirable = false;
        static constexpr bool is_preferable = false;
  
        using polymorphic_query_result_type = guarantee;
  
        template<class Executor>
          static constexpr bool static_query_v
            = Executor::query(blocking_t());
  
      constexpr guarantee value() const { return guarantee_; }

      private:
        guarantee guarantee_; // exposition only
    };

An advantage of this approach is that it would reduce the number of `.query()` overloads executors need to implement. Instead of implementing three different overloads for the blocking properties, only one would be needed.

Does this make sense, and if so, is it compatible with P0443R5's properties design?

jaredhoberock:
Think there may be a problem with such a scheme's interaction with `require()`.

We'd want to allow `ex.require(always_blocking)`, `ex.require(never_blocking)`, and `ex.require(possibly_blocking)` to return different types of executors. I don't see a way to do that with this approach.

AerialMantis:
I really like this idea, I was thinking of something along the same lines, but I came to the same conclusion that there would be no way to differentiate between the overloads of `require` or `query` for the properties in a mutually exclusive group. So this would mean that an implementation would have to support all or none of a particular group of properties.

I think making each property within a group a specialisation of a single template would be a nice way to enforce the properties as mutually exclusive. Another benefit of this is that you could have an overload of `query` for the `void` specialisation which returns the enum value so you don't have to call `query` multiple times. Though I imagine we wouldn't want a `require` overload for the `void` specialisation.

jaredhoberock:
Perhaps something like the following could work.

[...]

PR #360, opened 2018-02-27
Behavioral properties reformulation

jaredhoberock:
This is a WIP which reformulates the behavioral properties to enforce mutual exclusion via the type system.

### DIFF
--- a/wording.md
+++ b/wording.md
@@ -521,9 +521,9 @@ template<class Executor>
 
 ### Behavioral properties
 
-Behavioral properties define a set of mutually-exclusive enumerator properties describing executor behavior.
+Behavioral properties define a set of mutually-exclusive nested properties describing executor behavior.
 
-Unless otherwise specified, behavioral property types `S`, their enumerator property types `S::E`*i*, and enumerator property objects `S::e`*i* conform to the following specification:
+Unless otherwise specified, behavioral property types `S`, their nested property types `S::N`*i*, and nested property objects `S::n`*i* conform to the following specification:
 
     struct S
     {
@@ -540,7 +540,7 @@ Unless otherwise specified, behavioral property types `S`, their enumerator prop
 
       constexpr S();
 
-      struct E1
+      struct N1
       {
         static constexpr bool is_requirable = true;
         static constexpr bool is_preferable = true;
@@ -550,16 +550,16 @@ Unless otherwise specified, behavioral property types `S`, their enumerator prop
           static constexpr S static_query_v
             = Executor::query(E1());
 
-        static constexpr S value() { return S(E1()); }
+        static constexpr S value() { return S(N1()); }
       };
 
-      static constexpr e1;
+      static constexpr n1;
 
-      constexpr S(const E1);
+      constexpr S(const N1);
 
       ...
 
-      struct EN
+      struct NN
       {
         static constexpr bool is_requirable = true;
         static constexpr bool is_preferable = true;
@@ -567,14 +567,14 @@ Unless otherwise specified, behavioral property types `S`, their enumerator prop
 
         template<class Executor>
           static constexpr S static_query_v
-            = Executor::query(EN());
+            = Executor::query(NN());
 
-        static constexpr S value() { return S(EN()); }
+        static constexpr S value() { return S(NN()); }
       };
 
-      static constexpr eN;
+      static constexpr nN;
 
-      constexpr S(const EN);
+      constexpr S(const NN);
     };
 
 Queries for the value of an executor's behavioral property shall not change between calls unless the executor is assigned another executor with a different value of that behavioral property.
@@ -600,9 +600,9 @@ bool operator==(const S& a, const S& b);
 
 The `blocking_t` property describes what guarantees executors provide about the blocking behavior of their execution functions.
 
-`blocking_t` provides property enumerator types and objects as described below.
+`blocking_t` provides nested property types and objects as described below.
 
-| Property Enumerator Type | Property Enumerator Object Name | Requirements |
+| Nested Property Type | Nested Property Object Name | Requirements |
 |--------------------------|------------------------|--------------|
 | `blocking_t::possibly_t` | `blocking_t::possibly` | A call to an executor's execution function may block pending completion of one or more of the execution agents created by that execution function. |
 | `blocking_t::always_t` | `blocking_t::always` | A call to an executor's execution function shall block until completion of all execution agents created by that execution function. |
@@ -666,9 +666,9 @@ template<class Executor>
 
 The `blocking_adaptation_t` property allows or disallows blocking or directionality adaptation via `execution::require`.
 
-`blocking_adaptation_t` provides property enumerator types and objects as described below.
+`blocking_adaptation_t` provides nested property types and objects as described below.
 
-| Property Enumerator Type | Property Enumerator Object Name | Requirements |
+| Nested Property Type | Nested Property Object Name | Requirements |
 |--------------------------|---------------------------------|--------------|
 | `blocking_adaptation_t::allowed_t` | `blocking_adaptation::allowed` | The `require` customization point may adapt the executor to add the `twoway_t` or `blocking_t::always_t` properties. |
 | `blocking_adaptation_t::disallowed_t` | `blocking_adaptation::disallowed` | The `require` customization point may not adapt the executor to add the `twoway_t` or `blocking_t::always_t` properties. |
@@ -698,9 +698,9 @@ template<class Executor>
 
 The `relationship_t` property allows users of executors to indicate that submitted tasks represent continuations.
 
-`relationship_t` provides property enumerator types and objects as indicated below.
+`relationship_t` provides nested property types and objects as indicated below.
 
-| Property Enumerator Type | Property Enumerator Object Name | Requirements |
+| Nested Property Type | Nested Property Object Name | Requirements |
 |--------------------------|---------------------------------|--------------|
 | `relationship_t::continuation_t` | `relationship_t::yes` | Function objects submitted through the executor represent continuations of the caller. If the caller is a lightweight execution agent managed by the executor or its associated execution context, the execution of the submitted function object may be deferred until the caller completes. |
 | `relationship_t::fork_t` | `relationship_t::no` | Function objects submitted through the executor do not represent continuations of the caller. |
@@ -709,9 +709,9 @@ The `relationship_t` property allows users of executors to indicate that submitt
 
 The `outstanding_work_t` property allows users of executors to indicate that task submission is likely in the future.
 
-`outstanding_work_t` provides property enumerator types and objects as indicated below.
+`outstanding_work_t` provides nested property types and objects as indicated below.
 
-| Property Enumerator Type| Property Enumerator Object Name | Requirements |
+| Nested Property Type| Nested Property Object Name | Requirements |
 |-------------------------|---------------------------------|--------------|
 | `outstanding_work_t::tracked_t` | `outstanding_work::tracked` | The existence of the executor object represents an indication of likely future submission of a function object. The executor or its associated execution context may choose to maintain execution resources in anticipation of this submission. |
 | `outstanding_work_t::untracked_t` | `outstanding_work::untracked` | The existence of the executor object does not indicate any likely future submission of a function object. |
@@ -722,9 +722,9 @@ The `outstanding_work_t` property allows users of executors to indicate that tas
 
 Bulk execution guarantee properties communicate the forward progress and ordering guarantees of execution agents with respect to other agents within the same bulk submission.
 
-`bulk_guarantee_t` provides property enumerator types and objects as indicated below.
+`bulk_guarantee_t` provides nested property types and objects as indicated below.
 
-| Property Enumerator Type | Property Enumerator Object Name | Requirements |
+| Nested Property Type | Nested Property Object Name | Requirements |
 |--------------------------|---------------------------------|--------------|
 | `bulk_guarantee_t::sequenced_t` | `bulk_guarantee_t::sequenced` | Execution agents within the same bulk execution may not be parallelized. |
 | `bulk_guarantee_t::parallel_t` | `bulk_guarantee_t::parallel` | Execution agents within the same bulk execution may be parallelized. |
@@ -742,9 +742,9 @@ Execution agents created by executors with the `bulk_guarantee_t::unsequenced_t`
 
 The `mapping_t` property describes what guarantees executors provide about the mapping of execution agents onto threads of execution.
 
-`mapping_t` provides property enumerator types and objects as indicated below.
+`mapping_t` provides nested property types and objects as indicated below.
 
-| Property Enumerator Type| Property Enumerator Object Name | Requirements |
+| Nested Property Type| Nested Property Object Name | Requirements |
 |-------------------------|---------------------------------|--------------|
 | `mapping_t::other_t` | `mapping::other` | Mapping of each execution agent created by the executor is implementation-defined. |
 | `mapping_t::thread_t` | `mapping::thread` | Execution agents created by the executor are mapped onto threads of execution. |
--- a/wording.md
+++ b/wording.md
@@ -690,8 +690,8 @@ The `relationship_t` property allows users of executors to indicate that submitt
 
 | Nested Property Type | Nested Property Object Name | Requirements |
 |--------------------------|---------------------------------|--------------|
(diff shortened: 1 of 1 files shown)


jaredhoberock:
Each of these sets of properties include these riders at the end of their specifications:

> The value returned from `execution::query(e, p)` shall not change between calls unless `e` is assigned another executor which has a different value for `p`. The value returned from `execution::query(e, p1)` and a subsequent call `execution::query(e1, p1)` where:
> * `p1` is `execution::blocking.possibly`, `execution::blocking.always` or `execution::blocking.never`, and
> * `e1` is the result of `execution::require(e, p2)` or `execution::prefer(e, p2)`,
> shall compare equal unless:
> * `p2` is `execution::blocking.possibly`, `execution::blocking.always` or `execution::blocking.never`, and
> * `p1` and `p2` are different types.

We should probably hoist this common language into the common specification shared by all of them.

jaredhoberock:
Some of these properties could probably use a bikeshed now that these names have been rearranged.

For example, these names seem a bit awkward to me:

    `continuation.yes`
    `continuation.no`
    `outstanding_work.yes`
    `outstanding_work.no`
    `bulk_execution_guarantee.sequenced`
    `bulk_execution_guarantee.parallel`
    `bulk_execution_guarantee.unsequenced`

Additionally, I think we could probably nest the bulk execution guarantees inside the `bulk_t` property, since guarantees only apply to bulk executors. Doing so seems logical to me and would yield more reasonable names:

    `bulk`
    `bulk.sequenced`
    `bulk.parallel`
    `bulk.unsequenced`

chriskohlhoff:
> Some of these properties could probably use a bikeshed now that these names have been rearranged.

How about:

    outstanding_work.tracked
    outstanding_work.untracked

> Additionally, I think we could probably nest the bulk execution guarantees inside the bulk_t property, since these guarantees apply only to bulk executors. Doing so seems logical to me and would yield more reasonable names:

I suspect this gives problems for `query(ex, bulk_t)`, since that already has another static meaning.

jaredhoberock:
> I suspect this gives problems for query(ex, bulk_t), since that already has another static meaning.

`bulk_t::static_query_v<Executor>` returns true or false depending on the existence of a bulk execution function. I was thinking it could instead behave like the following:

    * ill-formed if no bulk execution functions exist
    * return `bulk` if a bulk execution function exists but `Executor::query(bulk)` is ill-formed
    * return `Executor::query(bulk)` if a bulk execution function exists and that expression is well-formed

Is there some problem that would interfere with that behavior? Perhaps something to do with the polymorphic executor?

chriskohlhoff:
> Is there some problem that would interfere with that behavior? Perhaps something to do with the polymorphic executor?

Possibly. I think we'd need a value that represents "not bulk", which is currently covered by the polymorphic executor's `query` function returning false. Feels a little like we're forcing two properties together in that case.

chriskohlhoff:
I think it may also interfere with the `require` customization point's automatic "identity" transformation, as that requires the query to be a constant expression. If an executor supports different bulk guarantee modes at runtime it can never be a constant expression, and the identity bullet point can't apply.

chriskohlhoff:
> Some of these properties could probably use a bikeshed now that these names have been rearranged.

For continuation/not_continuation, perhaps something like:

    relationship.continuation
    relationship.fork

jaredhoberock:
@chriskohlhoff @AerialMantis @ccmysen These changes should be ready for review.

wording.md
```diff
@@ -53,57 +53,53 @@ namespace execution {
...
-  struct never_blocking_t;
+  struct blocking_t;
 
-  constexpr possibly_blocking_t possibly_blocking;
-  constexpr always_blocking_t always_blocking;
-  constexpr never_blocking_t never_blocking;
+  constexpr blocking_t blocking;
+  constexpr blocking_t::possibly_t blocking_t::possibly;
+  constexpr blocking_t::always_t blocking_t::always;
+  constexpr blocking_t::never_t blocking_t::never;
```
chriskohlhoff: I don't think these nested constants belong in the header synopsis. It is sufficient to have them in the individual `blocking_t` class's synopsis.

wording.md
```diff
@@ -539,235 +535,242 @@ template<class Executor>
 
 ### Behavioral properties
 
-Unless otherwise specified, behavioral properties conform to the following specification:
+Behavioral properties define a set of mutually-exclusive enumerator properties describing executor behavior.
```
chriskohlhoff: I think the adaptable blocking property might be an exception to this statement.

wording.md
```diff
@@ -539,235 +535,242 @@ template<class Executor>
...
+
+Unless otherwise specified, behavioral property types `S`, their enumerator property types `S::E`*i*, and enumerator property objects `S::e`*i* conform to the following specification:
```
chriskohlhoff: Although they are designed to look and feel a lot like enumerators, the word does have a specific meaning within the language and these are not exactly that. Perhaps better just to say "nested".

wording.md
```diff
@@ -539,235 +535,242 @@ template<class Executor>
...
+
+The value returned from `execution::query(e1, p1)` and a subsequent call `execution::query(e1, p1)`, where
+
+* `p1` is an instance of `S` or `S::E`*i*, and
+* `e2` is the result of `execution::require(e1, p2)` or `execution::prefer(e1, p2)`,
+
+shall compare equal unless
+
+* `p2` is an instance of `S::E`*i*, and
 * `p1` and `p2` are different types.
```
chriskohlhoff: Need to say somewhere that `S()` is a distinct value to every `S::E`*i*.

chriskohlhoff: Should also say that every `S::E`i is distinct. (And by distinct I mean they all compare unequal.)

wording.md
```diff
@@ -539,235 +535,242 @@ template<class Executor>
...
+          static constexpr S static_query_v
+            = Executor::query(EN());
+
+        static constexpr S value() { return S(EN()); }
+      };
+
+      static constexpr eN;
+
+      constexpr S(const EN);
+    };
```
chriskohlhoff: Should this also include the `query` friend customization to enable automatic querying from any type. I.e. from https://github.com/executors/executors/issues/347#issuecomment-370104448:

    template<class Executor, class Property, class = typename std::enable_if<std::is_same<Property, blocking_t>::value>::type>
    friend constexpr auto query(const Executor& ex, const Property& p)
      noexcept(noexcept(execution::query(ex, std::declval<const query_helper<Executor>>())))
      -> decltype(execution::query(ex, std::declval<const query_helper<Executor>>()))
    {
      return execution::query(ex, query_helper<Executor>());
    }

but presumably worded with a "Shall not participate ...".

wording.md
```diff
@@ -539,235 +535,242 @@ template<class Executor>
...
+    struct possibly_t
     {
       template<class Executor>
-        static constexpr bool static_query_v
-          = see-below;
+      static constexpr blocking_t static_query_v
+        = see-below;
+
+      template<class Executor>
+        friend constexpr blocking_t query(const Executor& ex, blocking_t::possibly_t);
```
chriskohlhoff: I think I should have removed this as part of #362, as the static query covers it. Also below at line 650.

wording.md
```diff
@@ -539,235 +535,242 @@ template<class Executor>
...
+      struct E1
+      {
+        static constexpr bool is_requirable = true;
+        static constexpr bool is_preferable = true;
+        using polymorphic_query_result_type = S;
 
-    struct possibly_blocking_t;
-    struct always_blocking_t;
-    struct never_blocking_t;
+        template<class Executor>
```
AerialMantis: This may be tangential to the MR, but do we require `Executor::query(E1())` to be well-formed if `E1::static_query_v` is provided?

wording.md
```diff
@@ -539,235 +535,242 @@ template<class Executor>
...
+bool operator==(const S& a, const S& b);
+```
+
+*Returns:* `true` if `a` and `b` were constructed from the same constructor; `false`, otherwise.
+
+#### Blocking properties
+
+The `blocking_t` property describes what guarantees executors provide about the blocking behavior of their execution functions.
+
+`blocking_t` provides property enumerator types and objects as described below.
```
AerialMantis: If we decide to describe these as nested types rather than enumerated types as mentioned above, we may want to change this here as well.

wording.md
```diff
@@ -539,235 +535,242 @@ template<class Executor>
...
-    struct always_blocking_t
+    struct always_t
     {
       template<class Executor>
-        friend see-below require(Executor ex, always_blocking_t);
+        friend see-below require(Executor ex, blocking_t::always_t);
     };
 
-If the executor has the `adaptable_blocking_t` property, this customization uses an adapter to implement the `always_blocking_t` property.
+If the executor has the `blocking_t::adaptable_t` property, this customization uses an adapter to implement the `blocking_t::always_t` property.
```
chriskohlhoff: Should be `blocking_adaption_t::allowed_t`

wording.md
chriskohlhoff: Scratch that, i see you have also reformulated it as allowed/disallowed.

wording.md
```diff
@@ -539,235 +535,242 @@ template<class Executor>
...
 
-The value returned from `execution::query(e, p)` shall not change between calls unless `e` is assigned another executor which has a different value for `p`. The value returned from `execution::query(e, p1)` and a subsequent call `execution::query(e1, p1)` where:
-* `p1` is `execution::continuation` or `execution::not_continuation`, and
-* `e1` is the result of `execution::require(e, p2)` or `execution::prefer(e, p2)`,
-shall compare equal unless:
-* `p2` is `execution::continuation` or `execution::not_continuation`, and
-* `p1` and `p2` are different types.
+| Property Enumerator Type | Property Enumerator Object Name | Requirements |
+|--------------------------|---------------------------------|--------------|
+| `relationship_t::continuation_t` | `relationship_t::yes` | Function objects submitted through the executor represent continuations of the caller. If the caller is a lightweight execution agent managed by the executor or its associated execution context, the execution of the submitted function object may be deferred until the caller completes. |
```
chriskohlhoff: `yes` -> `continuation`

wording.md
```diff
@@ -539,235 +535,242 @@ template<class Executor>
...
+| `relationship_t::fork_t` | `relationship_t::no` | Function objects submitted through the executor do not represent continuations of the caller. |
```
chriskohlhoff: `no` -> `fork`.

wording.md
```diff
@@ -1358,7 +1355,7 @@ The polymorphic `executor` wrapper should be able to simply swap in, so that we
         executor<
           execution::single,
           execution::oneway,
-          execution::outstanding_work> ex,
+          execution::outstanding_work.tracked> ex,
         std::function<void()> callback)
```
chriskohlhoff: Existing mistake here in preceding template parameter list: missing `_t` suffixes. Should be `execution::outstanding_work_t::tracked_t` for example.

wording.md
```diff
@@ -1374,22 +1371,22 @@ The polymorphic `executor` wrapper should be able to simply swap in, so that we
...
 
 The `prefer_only` adapter addresses this by turning off the `is_requirable` attribute for a specific property. It would be used in the above example as follows:
 
     void do_async_work(
         executor<
           execution::single,
           execution::oneway,
-          prefer_only<execution::outstanding_work>> ex,
+          prefer_only<execution::outstanding_work.tracked>> ex,
         std::function<void()> callback)
```
chriskohlhoff: Existing mistake here in preceding template parameter list: missing `_t` suffixes. Should be `execution::outstanding_work_t::tracked_t` for example.

wording.md
AerialMantis: Are `relationship_t::yes` and `relationship_t::no` aliases to `relationship_t::continuation_t` and `relationship_t::fork_t`, I don't see them mentioned in the synopsis.

wording.md
```diff
@@ -539,235 +535,242 @@ template<class Executor>
...
-| `new_thread_execution_mapping_t` | Each execution agent created by the executor is mapped onto a new thread of execution. |
+The `mapping_t` property describes what guarantees executors provide about the mapping of execution agents onto threads of execution.
 
-The `other_execution_mapping_t`, `thread_execution_mapping_t` and `new_thread_execution_mapping_t` properties are mutually exclusive.
+`mapping_t` provides property enumerator types and objects as indicated below.
 
-[*Note:* The guarantees of `other_execution_mapping_t`, `thread_execution_mapping_t` and `new_thread_execution_mapping_t` implies the relationship: `other_execution_mapping_t < thread_execution_mapping_t < new_thread_execution_mapping_t` *--end note*]
+| Property Enumerator Type| Property Enumerator Object Name | Requirements |
+|-------------------------|---------------------------------|--------------|
+| `mapping_t::other_t` | `mapping::other` | Mapping of each execution agent created by the executor is implementation-defined. |
```
AerialMantis: The previous enumerator object names have the `_t` suffix, but these ones don't.

wording.md
```diff
@@ -53,57 +53,53 @@ namespace execution {
...
 
-  struct bulk_sequenced_execution_t;
-  struct bulk_parallel_execution_t;
-  struct bulk_unsequenced_execution_t;
+  struct bulk_guarantee_t;
 
-  constexpr bulk_sequenced_execution_t bulk_sequenced_execution;
-  constexpr bulk_parallel_execution_t bulk_parallel_execution;
-  constexpr bulk_unsequenced_execution_t bulk_unsequenced_execution;
+  constexpr bulk_guarantee_t bulk_guarantee_t;
```
AerialMantis: Should this be `bulk_guarantee_t  bulk_guarantee`?

wording.md
```diff
@@ -53,57 +53,53 @@ namespace execution {
...
 
   // Properties to indicate if submitted tasks represent continuations:
 
-  struct not_continuation_t;
-  struct continuation_t;
+  struct relationship_t;
 
-  constexpr not_continuation_t not_continuation;
-  constexpr continuation_t continuation;
+  constexpr relationship_t relationship_t;
```
AerialMantis: Should this be `constexpr relationship_t relationship;`?

wording.md
```diff
@@ -53,57 +53,53 @@ namespace execution {
...
   // Properties to allow adaptation of blocking and directionality:
 
-  struct adaptable_blocking_t;
-  struct not_adaptable_blocking_t;
+  struct blocking_adaptation_t;
 
-  constexpr adaptable_blocking_t adaptable_blocking;
-  constexpr not_adaptable_blocking_t not_adaptable_blocking;
+  constexpr blocking_adaptation_t::allowed_t blocking_adaptation_t::allowed;
+  constexpr blocking_adaptation_t::allowed_t blocking_adaptation_t::disallowed;
```
AerialMantis: Should this be `constexpr blocking_adaptation_t::disallowed_t blocking_adaptation_t::disallowed;`

wording.md
AerialMantis: Okay from seeing the other properties I think I understand what this means now, so `relationship_t::yes` and `relationship_t::no` are the instance names, so you would call `require(e, relationship.yes)`, is that correct? If so I think, unless I'm misunderstanding this, there may be a confusion between `relationship.continuation` / `relationship.fork` and `continuation.yes` / `continuation.no`.

jaredhoberock: I think the `relationship.yes` and `relationship.no` are just typos. The names ChrisK suggested for these were `relationship.continuation` and `relationship.fork`.

wording.md
jaredhoberock: I'll add it. In the example, why is the type of `Property` deduced for `query()` instead of simply being `blocking_t`?

wording.md
jaredhoberock: I think we are going to introduce `S::query(E1())` per ChrisK's comment up the thread.

wording.md
chriskohlhoff: Yeah it's deduced to prevent a similar issue to #352.

jaredhoberock: Thanks. I'm trying to figure out when this friend `query()` should be in the overload set due to the example code's use of `query_helper`. I think it's something like:

1. `Property` is `S`
2. `can_query_v<Executor, S::Ni>` for at least one `S::Ni`?

Does that sound right to you?

chriskohlhoff: Yes, looks right.

It might also be worth specifying which one it chooses to return, e.g. the first *i* for which `can_query_v<Executor, S::Ni>` is true.

jaredhoberock:
Thanks guys, I think your feedback is now incorporated into the PR. I'm going to go ahead and merge so I can more easily fix some typos and then I'll post the draft paper to the wiki.

PR #384, opened 2018-04-16
PromiseExecutor and ReadyFuture

LeeHowes:
Details the requirements for ReadyFuture and PromiseExecutor.

PromiseExecutor exposes get_promise to provide a means to transition from one executor to another executor which know nothing about each other.

Defines Promise in terms of ReadyFuture as a concrete type to parameterise tasks.

### DIFF
--- a/wording.md
+++ b/wording.md
@@ -38,10 +38,12 @@ namespace execution {
   struct oneway_t;
   struct twoway_t;
   struct then_t;
+  struct promise_t;
 
   constexpr oneway_t oneway;
   constexpr twoway_t twoway;
   constexpr then_t then;
+  constexpr promise_t promise;
 
   // Cardinality properties:
 
@@ -99,6 +101,7 @@ namespace execution {
   template<class Executor> struct is_oneway_executor;
   template<class Executor> struct is_twoway_executor;
   template<class Executor> struct is_then_executor;
+  template<class Executor> struct is_promise_executor;
   template<class Executor> struct is_bulk_oneway_executor;
   template<class Executor> struct is_bulk_twoway_executor;
   template<class Executor> struct is_bulk_then_executor;
@@ -106,6 +109,7 @@ namespace execution {
   template<class Executor> constexpr bool is_oneway_executor_v = is_oneway_executor<Executor>::value;
   template<class Executor> constexpr bool is_twoway_executor_v = is_twoway_executor<Executor>::value;
   template<class Executor> constexpr bool is_then_executor_v = is_then_executor<Executor>::value;
+  template<class Executor> constexpr bool is_promise_executor_v = is_promise_executor<Executor>::value;
   template<class Executor> constexpr bool is_bulk_oneway_executor_v = is_bulk_oneway_executor<Executor>::value;
   template<class Executor> constexpr bool is_bulk_twoway_executor_v = is_bulk_twoway_executor<Executor>::value;
   template<class Executor> constexpr bool is_bulk_then_executor_v = is_bulk_then_executor<Executor>::value;
@@ -158,6 +162,10 @@ A type `F` meets the `Future` requirements for some value type `T` if `F` is `st
 
 Forward progress guarantees are a property of the concrete `Future` type. [*Note:* `std::experimental::future<T>::wait()` blocks with forward progress guarantee delegation until the shared state is ready. *--end note*]
 
+### `Promise` requirements
+
+A type `P` meets the `Promise` requirements for some value `T` if an instance `p` of `P` is callable as `DECAY_COPY(std::forward<P>(p))(std::variant<T, std::exception_ptr>)` and if `P` has a method `get_future()` that returns a `Future` of type `T` that completes on a call to `p(v)` with the value or exception stored within `v`.
+
 ### `ProtoAllocator` requirements
 
 A type `A` meets the `ProtoAllocator` requirements if `A` is `CopyConstructible` (C++Std [copyconstructible]), `Destructible` (C++Std [destructible]), and `allocator_traits<A>::rebind_alloc<U>` meets the allocator requirements (C++Std [allocator.requirements]), where `U` is an object type. [*Note:* For example, `std::allocator<void>` meets the proto-allocator requirements but not the allocator requirements. *--end note*] No comparison operator, copy operation, move operation, or swap operation on these types shall exit via an exception.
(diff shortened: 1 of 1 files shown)


wording.md
```diff
@@ -158,6 +162,26 @@ A type `F` meets the `Future` requirements for some value type `T` if `F` is `st
...
+
+### `ReadyFuture` requirements
+
+A type `RFR` meets the `ReadyFuture` requirements for some value type `T` if an instance `rfr` of `RFR` satisfies the requirements in the table below.
+
+In the table below `RFR2` is a type that meet the `ReadyFuture` requirements, and may be distinct from `RFR`. `e` is a value that meets the `OneWayExecutor` or `PromiseExecutor` requirements.
+
+| Expression | Return Type | Operational semantics |
+|------------|-------------|---------------------- |
+| `rfr(const RFR2&)` | None | Copy-construction from `RFR` or from another `ReadyFuture` type if `T` is copy-constructible. |
```
LeeHowes: I actually think this should only be potentially copy- and move-constructible from itself. Not from all ReadyFutures. Instead Promise should not be a simple ReadyFuture taking callback but instead be a pair of set_value and set_exception. That way ReadyFuture really is a pure read type that does not have to leave callbacks its own executor. Interfacing between executors has an explicit value/exception split.

wording.md
```diff
@@ -268,15 +297,35 @@ In the Table below,
...
+
+  * `x` denotes a (possibly const) executor object of type `X`,
+  * `n` denotes a shape object whose type is `executor_shape_t<X>`,
+  * `fut` denotes a future object satisfying the Future requirements,
+  * `rf` denotes a `CopyConstructible` function object with zero arguments whose result type is `R`,
+  * `sf` denotes a `CopyConstructible` function object with zero arguments whose result type is `S`,
+
+| Expression | Return Type | Operational semantics |
+|------------|-------------|---------------------- |
+| `x.get_promise<T>` | A pair of a value of a type that satisfies the `Promise` requirements and a value of a type that satisfies the `Future` requirements. | Returns a `Promise` `p` and a `Future` f such that calling `p.set_value(v)` completes `f` with a value and calling `p.set_exception(e)` completes `f` with an exception. |
```
LeeHowes: Leaving this for now because it doesn't feel like a query as-such. I think a customisation point might be the right thing here.

wording.md
```diff
@@ -99,22 +101,26 @@ namespace execution {
...
   template<class Executor> constexpr bool is_bulk_twoway_executor_v = is_bulk_twoway_executor<Executor>::value;
   template<class Executor> constexpr bool is_bulk_then_executor_v = is_bulk_then_executor<Executor>::value;
 
   template<class Executor, class T> struct executor_future;
+  template<class Executor, class T> struct executor_ready_future;
   template<class Executor> struct executor_shape;
   template<class Executor> struct executor_index;
 
   template<class Executor, class T> using executor_future_t = typename executor_future<Executor, T>::type;
+  template<class Executor, class T> using executor_ready_future_t = typename executor_ready_future<Executor, T>::type;
```
LeeHowes: What I would like here would be for us to optionally allow:
`template<class Executor, class T, class ExceptionTypes...> using executor_ready_future_t = typename executor_ready_future<Executor, T, ExceptionTypes...>::type;`

Such that we can provide that exception filter list later. I don't know if that becomes a break to add an optional list of types. It could also of course be achieved by simply adding a new type later:
`template<class Executor, class T, class ExceptionTypes...> using executor_filtered_ready_future_t = typename executor_filtered_ready_future<Executor, T, ExceptionTypes...>::type;`

wording.md
LeeHowes: Note that it returns a pair. The reason it returns a pair is that I do not want the Promise concept to require a get_future method, because that restricts implementations.

Instead here we return a promise and a future that are paired. The future must be passable to then_execute and bulk_then_execute. The promise can be satisfied by a caller to trigger such work.

We would probably want some wording in here somewhere to require that both executor_future_t and the future returned here be compatible with then_execute. They probably should not have to be the same type.

chriskohlhoff:
After pondering it for the week, I am of the opinion that a property with a corresponding query is the right mechanism for this. A query will:

* Automatically work with adapters
* Automatically work with the polymorphic wrapper

as well as decoupling the core executors from a the promise concept, and allowing refinement of the futures framework independent of core executors.

There are a few options for how this query could look. One is a direct query for a new promise from an executor:

    template <T>
    struct promise_t
    {
      // ...
    };

    auto prom = execution::query(ex, execution::promise_t<foo>());

Another is a query that returns a future/promise pair:

    template <T>
    struct futures_contract_t
    {
      // ...
    };

    auto pair = execution::query(ex, execution::futures_contract_t<foo>());
    auto fut = pair.first;
    auto prom = pair.first;

Yet another is a query that returns a kind of factory object that allows creation of potentially different kinds of futures:

    template <T>
    struct futures_factory_t
    {
      // ...
    };

    auto factory = execution::query(ex, execution::futures_factory);
    auto prom = factory.make_promise<foo>();
    auto fut = factory.make_ready_future<int>(123);

Whichever type of property is used, it is straightforward to use the properties mechanism to provide a default query for common scenarios, e.g. for executors where executor_future_t yields a std::promise, the execution::promise_t property can have a default query that creates a new std::promise object.

Using the properties gives as the flexibility to add additional ones as the futures framework grows, e.g. why not also have:

    auto fut = execution::query(ex, execution::ready_future(123));

wording.md
AerialMantis: Perhaps we should name this `make_promise` rather than `get_promise`, as from my understanding it's creating a new promise future pair.

LeeHowes:
Thanks Chris. I agree, and as you can see from my notes in the code here I prefer the approach of returning a pair because it keeps the Promise concept simple and more flexible. I lean towards a query that returns a pair at this time.

wording.md
LeeHowes: You are right. Chris suggested that too and I effectively ignored his suggestion pending his thoughts on whether a query worked better anyway. I think it does - we can call the query something along the lines of make promise.

AerialMantis: Yeah, that could work. My only concern with that is it would go against the precedent in the paper just now where the value returned from a query doesn't change between calls, though we could always have a category of property which essentially perform make-like behaviour and doesn't conform to this.

LeeHowes:
Notes: Primary Try type part of this agreed during call to make a single struct with operator()(value) and operator()(tag_type, exception_ptr). Jared is drafting.

Query part will be proposed as a separate paper on top of the executors proposal.

PR #386, opened 2018-04-19
Give static_query_v a default value, when appropriate

### LINKED ISSUE #380: bulk_guarantee.unsequenced hasn't had the full static_query_v treatment
`bulk_guarantee.unsequenced` is analogous to `blocking.possibly` in the sense that if the executor doesn't provide an explicit bulk guarantee, then the client should assume that groups of agents will execute in an unsequenced manner. In other words, executors should have the `bulk_guarantee.unsequenced` property by default.

To fix this, we should specify `unsequenced_t` as follows:

> In addition to conforming to the above specification, the blocking_t::possibly_t property provides the following customization:
> 
> ```
> struct unsequenced_t
> {
>   template<class Executor>
>   static constexpr bulk_guarantee_t static_query_v
>     = see-below;
> 
>   template<class Executor>
>     friend constexpr bulk_guarantee_t query(const Executor& ex, bulk_guarantee_t::unsequenced_t);
> };
> ```
> These members automatically enable the bulk_guarantee_t::unsequenced_t property for all executors that do not otherwise support the `bulk_guarantee_t::sequenced_t` or `bulk_guarantee_t::parallel_t` properties. [Note: That is, all executors are treated as bulk unsequenced unless otherwise specified. --end note]
> 
> The `static_query_v` member is:
> 
> `Executor::query(bulk_guarantee_t::unsequenced_t{})` if that is a well-formed constant expression.
> ill-formed if `declval<Executor>.query(bulk_guarantee_t::unsequenced_t{})` is well-formed;
> ill-formed if `can_query_v<Executor, bulk_guarantee_t::sequenced_t>` is true;
> ill-formed if `can_query_v<Executor, bulk_guarantee_t::parallel_t>` is true;
> otherwise `bulk_guarantee.unsequenced`.
> 
> ```
> template<class Executor>
>   friend constexpr blocking_t query(const Executor& ex, bulk_guarantee_t::unsequenced_t);
> ```
> 
> Returns: `bulk_guarantee.unsequenced`.
> 
> Remarks: This function shall not participate in overload resolution unless `can_query_v<Executor, bulk_guarantee_t::sequenced_t> || can_query_v<Executor, bulk_guarantee_t::parallel_t>` is false.

jaredhoberock:
This change fixes `static_query_v` to result in a default value for behavioral properties. Since the default property value is the first nested property type, we should probably audit the behavioral property tables to ensure that the default value is listed first.

Fixes #380

### DIFF
--- a/wording.md
+++ b/wording.md
@@ -629,6 +629,7 @@ The value of the expression `S::N`*i*`::static_query_v<Executor>` is
 The value of the expression `S::static_query_v<Executor>` is
 
 * `Executor::query(S())`, if that expression is a well-formed constant expression;
+* otherwise, ill-formed if `declval<Executor>().query(S())` is well-formed;
 * otherwise, `S::N`*i*`::static_query_v<Executor>` for the least *i* `<= N` for which this expression is a well-formed constant expression;
 * otherwise, ill-formed.
 
--- a/wording.md
+++ b/wording.md
@@ -617,23 +617,28 @@ shall compare equal unless
 * `p2` is an instance of `S::E`*i*, and
 * `p1` and `p2` are different types.
 
-The value of the expression `S::N`*i*`::static_query_v<Executor>` is
+The value of the expression `S::N1::static_query_v<Executor>` is
 
-* `Executor::query(S::N`*i*`())`, if that expression is a well-formed constant expression;
-* ill-formed if `declval<Executor>().query(S::N`*i*`())` is well-formed;
-* ill-formed if `can_query_v<Executor, S::N`*i*`>` is `true` for all `1 <` *i* `<= N`;
+* `Executor::query(S::N1())`, if that expression is a well-formed expression;
+* ill-formed if `declval<Executor>().query(S::N1())` is well-formed;
+* ill-formed if `can_query_v<Executor,S::N`*i*`>` is `true` for any `1 < ` *i* `<= N`;
 * otherwise `S::N1()`.
 
-[*Note:* These rules automatically enable the `S::N1` property by default for executors which do not provide a `query` function for properties `S::N`*i*`. *--end note*]
+[*Note:* These rules automatically enable the `S::N1` property by default for executors which do not provide a `query` function for properties `S::N`*i*. *--end note*]
+
+The value of the expression `S::N`*i*`::static_query_v<Executor>`, for all `1 < ` *i* `<= N`, is
+
+* `Executor::query(S::N`*i*`())`, if that expression is a well-formed constant expression;
+* otherwise ill-formed.
 
 The value of the expression `S::static_query_v<Executor>` is
 
 * `Executor::query(S())`, if that expression is a well-formed constant expression;
 * otherwise, ill-formed if `declval<Executor>().query(S())` is well-formed;
 * otherwise, `S::N`*i*`::static_query_v<Executor>` for the least *i* `<= N` for which this expression is a well-formed constant expression;
-* otherwise, ill-formed.
+* otherwise ill-formed.
 
-[*Note:* These rules automatically enable the `S::N1` property by default for executors which do not provide a `query` function for properties `S` or `S::N`*i*`. *--end note*]
+[*Note:* These rules automatically enable the `S::N1` property by default for executors which do not provide a `query` function for properties `S` or `S::N`*i*. *--end note*]
 
 Let *k* be the least value of *i* for which `can_query_v<Executor,S::N`*i*`>` is true, if such a value of *i* exists.
 
--- a/wording.md
+++ b/wording.md
@@ -664,27 +664,6 @@ The `blocking_t` property describes what guarantees executors provide about the 
 | `blocking_t::always_t` | `blocking_t::always` | A call to an executor's execution function shall block until completion of all execution agents created by that execution function. |
 | `blocking_t::never_t` | `blocking_t::never` | A call to an executor's execution function shall not block pending completion of the execution agents created by that execution function. |
 
-##### `blocking_t::possibly_t` customization points
-
-In addition to conforming to the above specification, the `blocking_t::possibly_t` property provides the following customization:
-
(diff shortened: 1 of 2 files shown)


jaredhoberock:
Questions for reviewers:

1. `blocking_t::always_t` has a `static_query_v` customization. Do the other default properties need a similar customization? If so, can we hoist this up into the general Behavioral Properties specification?

2. What should be the default value of `blocking_adaptation_t`? `blocking_adaptation.disallowed`, right?

3. What should be the default value of `relationship_t`?

4. What should be the default value of `outstanding_work_t`?

wording.md
```diff
@@ -619,11 +619,24 @@ shall compare equal unless
...
 ```
 template<class Executor>
-  friend constexpr S query(const Executor& ex, const Property& p) noexcept(noexcept(execution::query(ex, std::declval<const S::Nk>())));
+  static constexpr auto static_query_v = see-below;
 ```
 
+The value of the expression `S::static_query_v<Executor>` is
+
+* `Executor::query(S())`, if that expression is a well-formed constant expression.
+* Otherwise, `S::N1()`, if that expression is a well-formed constant expression, and if `can_query_v<Executor,S>` is false and if `can_query_v<Executor,S::N`*i*`>` is false for all *i*.
```
chriskohlhoff: should this be for all i > 1?

jaredhoberock: if `can_query_v<Executor,S::N1>` is true, that means that either `ex.query(S::N1())` or `query(ex, S::N1())` is well-formed. In such a case, we would not want to override that result with the default value.

Is that right, or have I got it wrong?

chriskohlhoff:
1. If we're going to say all enumerated properties have a default, then yes it should be hoisted into the common section.

2. disallowed

3. I think fork is appropriate, i.e. unless otherwise specified you're asking for a fork in the chained control flow.

4. I think untracked is appropriate and matches the existing uses in the network TS, i.e. you have to explicitly opt into work tracking

What about thread mapping?

jaredhoberock:
Thread mapping is a good question. Even though `mapping.other` provides the least guarantees, I think we should probably assume `mapping.thread` by default for C++.

wording.md
chriskohlhoff: Sorry, I thought this bit was in S::N1::static_query_v. However, doesn't this prevent:

    query(ex, S())

from working when the executor only implements query for one of S::N*i*?

My implementation uses integers to make it clearer how it follows the spec:

https://github.com/chriskohlhoff/asio/blob/unified-executors/asio/include/asio/execution/detail/enumeration.hpp

For S::static_query_v this implementation performs:

* `Executor::query(S())`, if that expression is a well-formed constant expression.
* Otherwise, `S::Ni()`, if that expression is a well-formed constant expression, for the lowest numbered *i*.

And, for enumerations having a default, S::N1 (actually N0 in my implementation) has the static_query_v that checks !can_query_v for all *i* > 1.

jaredhoberock:
@chriskohlhoff I'm finished with these changes. Do they look correct to you?

wording.md
```diff
@@ -617,13 +617,31 @@ shall compare equal unless
 * `p2` is an instance of `S::E`*i*, and
 * `p1` and `p2` are different types.
 
+The value of the expression `S::N`*i*`::static_query_v<Executor>` is
+
+* `Executor::query(S::N`*i*`())`, if that expression is a well-formed constant expression;
+* ill-formed if `declval<Executor>().query(S::N`*i*`())` is well-formed;
+* ill-formed if `can_query_v<Executor, S::N`*i*`>` is `true` for all `1 <` *i* `<= N`;
```
chriskohlhoff: This formulation isn't quite the same as what I've done in the asio branch, because it doesn't distinguish between the specification for the default and non-default enumerators.

I've been trying to update the implementation to match the above bullet points but I can't get it to work and suspect it isn't possible. The problem seems to be that the use of can_query_v<> is circular, i.e. every S::N`i` uses can_query_v for all other values of `i`.

My existing implementation seems to work because it's a strictly acyclic graph, i.e. property 1 tests can_query_v for properties 2..N, but not the other way around. I think the way to specify it would be something like:

The value of the expression `S::N1::static_query_v<Executor>` is:

* `Executor::query(S::N1())`, if that expression is a well-formed constant expression;
* ill-formed if `declval<Executor>().query(S::N`*i*`())` is well-formed;
* ill-formed if `can_query_v<Executor, S::N`*i*`>` is `true` for all `1 <` *i* `<= N`;
* otherwise `S::N1()`.

The value of the expression `S::N`*i*`::static_query_v<Executor>` is:

* `Executor::query(S::N`*i*`())`, if that expression is a well-formed constant expression;
* otherwise ill-formed.

wording.md
```diff
@@ -617,13 +617,31 @@ shall compare equal unless
...
+
+* `Executor::query(S())`, if that expression is a well-formed constant expression;
+* otherwise, ill-formed if `declval<Executor>().query(S())` is well-formed;
+* otherwise, `S::N`*i*`::static_query_v<Executor>` for the least *i* `<= N` for which this expression is a well-formed constant expression;
+* otherwise, ill-formed.
+
+[*Note:* These rules automatically enable the `S::N1` property by default for executors which do not provide a `query` function for properties `S` or `S::N`*i*`. *--end note*]
+
+Let *k* be the least value of *i* for which `can_query_v<Executor,S::N`*i*`>` is true, if such a value of *i* exists.
+
```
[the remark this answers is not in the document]
chriskohlhoff: I tried this change but it prevented my test cases from compiling. I think safer to stay with only testing for the member function form (member function == executor supports the property "natively").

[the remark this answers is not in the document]
jaredhoberock: Sounds good.

wording.md
jaredhoberock: Thanks, updated.

jaredhoberock:
Thanks, merged.

PR #387, opened 2018-04-23
Revise then_execute & bulk_then_execute's operational semantics to allow exception handling

### LINKED ISSUE #383: .bulk_then_execute expects its user function to receive a future parameter
The current proposal describes `.bulk_then_execute` this way:

> If `R` is non-`void`, invokes `rf()` at most once on an unspecified execution agent to produce the value `r`. Invokes `sf()` at most once on an unspecified execution agent to produce the value `s`. When `fut` is ready, creates a group of execution agents of shape `n` which invokes `DECAY_COPY( std::forward<F>(f))(i, fut, r, s)` if `R` is non-`void`, and otherwise invokes `DECAY_COPY( std::forward<F>(f))(i, fut, s)`, at most once for each value of `i` in the range `[0,n)`, with the call to `DECAY_COPY` being evaluated in the thread that called `bulk_then_execute`.

However, in R3, `.bulk_then_execute`'s operational semantics include:

> If `R` is non-`void`, invokes `rf()` on an unspecified execution agent to produce the value `r`. Invokes `sf()` on an unspecified execution agent to produce the value `s`. When `pred` is ready, creates a group of execution agents of shape `n` which invokes `DECAY_COPY( std::forward<F>(f))(i, pred, r, s)` if `R` is non-`void`, and otherwise invokes `DECAY_COPY( std::forward<F>(f))(i, pred, s)`, at most once for each value of `i` in the range `[0,n)`, with the call to `DECAY_COPY` being evaluated in the thread that called `bulk_twoway_execute` [sic].

In this text, `pred` is the value of the predecessor future.

The current text requires `f` to receive the future, rather than the future's ready value, as a parameter. The problem is that it this implies that each execution agent in the group must call `fut.get()` to unwrap its value. If multiple agents do this in parallel, they will race.

We could revert to the older semantics, but this would create an inconsistency with `.then_execute`'s expectations.

jaredhoberock:
This is a WIP to resolve #383. I'd like to get the simpler `then_execute` semantics in place, and then generalize those semantics to `bulk_then_execute`.


wording.md
```diff
@@ -204,11 +204,11 @@ The `ThenExecutor` requirements specify requirements for executors which create
...
 
-In the Table below, `x` denotes a (possibly const) executor object of type `X`, `fut` denotes a future object satisfying the Future requirements, `f` denotes a function object of type `F&&` callable as `DECAY_COPY(std::forward<F>(f))(fut)` and where `decay_t<F>` satisfies the `MoveConstructible` requirements, and `R` denotes the type of the expression `DECAY_COPY(std::forward<F>(f))(fut)`.
+In the Table below, `x` denotes a (possibly const) executor object of type `X`, `fut` denotes a future object satisfying the Future requirements, `val` denotes the object stored in `fut`'s associated shared state when it becomes ready and nonexceptional, `e` denotes the object stored in `fut`'s associated shared state when it becomes ready and exceptional, `f` denotes a function object of type `F&&` callable as `DECAY_COPY(std::forward<F>(f))(val)` or `DECAY_COPY(std::forward<F>(f))(exception_arg, e)` and where `decay_t<F>` satisfies the `MoveConstructible` requirements, and `R` denotes the type of the expression `DECAY_COPY(std::forward<F>(f))(val)`.
```
LeeHowes: Shouldn't we be moving val here?

What does this mean for a Future<void> as the input future?

jaredhoberock: I don't think we should move `val` into the invocation of `f` in `.then_execute()` because some types which can be stored in futures aren't movable. `.then()` could work differently, however.

> What does this mean for a Future as the input future?

What do you mean?

wording.md
```diff
@@ -204,11 +204,11 @@ The `ThenExecutor` requirements specify requirements for executors which create
...
 
-In the Table below, `x` denotes a (possibly const) executor object of type `X`, `fut` denotes a future object satisfying the Future requirements, `f` denotes a function object of type `F&&` callable as `DECAY_COPY(std::forward<F>(f))(fut)` and where `decay_t<F>` satisfies the `MoveConstructible` requirements, and `R` denotes the type of the expression `DECAY_COPY(std::forward<F>(f))(fut)`.
+In the Table below, `x` denotes a (possibly const) executor object of type `X`, `fut` denotes a future object satisfying the Future requirements, `val` denotes any object stored in `fut`'s associated shared state when it becomes ready and nonexceptional, `e` denotes the object stored in `fut`'s associated shared state when it becomes ready and exceptional, `f` denotes a function object of type `F&&` callable as `DECAY_COPY(std::forward<F>(f))(val)` or `DECAY_COPY(std::forward<F>(f))(exception_arg, e)` and where `decay_t<F>` satisfies the `MoveConstructible` requirements, and `R` denotes the type of the expression `DECAY_COPY(std::forward<F>(f))(val)`.
```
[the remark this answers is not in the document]
jaredhoberock: > if NORMAL and EXCEPTIONAL are both satisfied, both must return the same type.

The current text in the operational semantics says in this case that `decltype(EXCEPTIONAL)` shall be convertible to `R` (which is `decltype(NORMAL)`). Is that close enough?

> If only EXCEPTIONAL is satisfied, then the return type of the function must be the same as the type of val."

I had assumed that if only `EXCEPTIONAL` is well-formed, then the result of `f(exception_arg, e)` could be anything. What is the reasoning for wanting it to be the same type as `val`?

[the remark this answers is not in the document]
jaredhoberock: Got it. I'll update.

jaredhoberock:
@LeeHowes I've added revised semantics for `bulk_then_execute`, based on generalizing `then_execute`'s revision. Does it look correct to you?

wording_front_matter.md
```diff
@@ -53,6 +53,9 @@ Abstract:           This paper proposes a programming model for executors, which
 
 TODO
 
+* Revise `then_execute` and `bulk_then_execute`'s operational semantics to allow user functions to handle incoming exceptions thrown by preceding execution agents
```
LeeHowes: Out of date now?

LeeHowes:
I think that looks right based on what we discussed.

jaredhoberock:
Thanks Lee, merged.

PR #402, opened 2018-06-07
Introduce model

hcedwar:
Introduce front matter to model execution agent, execution context, execution resource, and executor.
Change requirement that executor execute functions immediately create execution agents to executor causes callable object to be invoked.

### DIFF
--- a/wording.md
+++ b/wording.md
@@ -22,15 +22,15 @@ The `exception_arg_t` struct is an empty structure type used as a unique type to
 
 ### General
 
-This Clause describes components supporting execution of a callable object [func.def].
+This Clause describes components supporting execution of function objects [function.objects].
 
-As defined in [thread.req.lockable.general]:
+*(The following definition appears in working draft N4762 [thread.req.lockable.general])*
 
-> "An *execution agent* is an entity such as a thread that may perform work in parallel with other execution agents. [*Note:* Implementations or users may introduce other kinds of agents such as processes or thread-pool tasks. *--end note*] The calling agent is determined by context; e.g., the calling thread that contains the call, and so on." 
+> An *execution agent* is an entity such as a thread that may perform work in parallel with other execution agents. [*Note:* Implementations or users may introduce other kinds of agents such as processes or thread-pool tasks. *--end note*] The calling agent is determined by context; e.g., the calling thread that contains the call, and so on. 
 
-An execution agent supports execution by invoking a callable within an *execution context* such as as the calling thread or thread-pool.  An *executor* submits a callable to an execution context to be invoked by an execution agent within an associated execution context. [*Note:* Invocation of the callable may be inlined such as when the execution context is the calling thread, or may be scheduled such as when the execution context is a thread-pool with task scheduler. *--end note*] An executor may submit a callable with *execution properties* that specify how the submission and invocation of the callable interacts with the submitting thread and execution context, including forward progress guarantees [intro.progress]. 
+An execution agent invokes a function object  within an *execution context* such as the calling thread or thread-pool.  An *executor* submits a function object to an execution context to be invoked by an execution agent within that execution context. [*Note:* Invocation of the function object may be inlined such as when the execution context is the calling thread, or may be scheduled such as when the execution context is a thread-pool with task scheduler. *--end note*] An executor may submit a function object with *execution properties* that specify how the submission and invocation of the function object interacts with the submitting thread and execution context, including forward progress guarantees [intro.progress]. 
 
-For the intent of this library and extensions to this library, the observable *lifetime of an execution agent* begins immediately before the callable is invoked and ends immediately after the callable completes, either normally or having thrown an exception.
+For the intent of this library and extensions to this library, the *lifetime of an execution agent* begins before the function object is invoked and ends after this invocation completes, either normally or having thrown an exception.
 
 
 ### Header `<execution>` synopsis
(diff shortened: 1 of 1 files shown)


wording.md
```diff
@@ -256,11 +272,11 @@ In the Table below,
...
+| `x.bulk_execute(f, n, sf)` | `void` | Causes `sf()` to be invoked at most once to produce the value `s`. Causes `DECAY_COPY( std::forward<F>(f))(i, s)` to be invoked at most once for each value of `i` in the range `[0,n)`, with the call to `DECAY_COPY` being evaluated in the thread that called `bulk_execute`. <br/> <br/> May block pending completion of one or more calls to `DECAY_COPY( std::forward<F>(f))(i, s)`. <br/> <br/> The invocation of `bulk_execute` synchronizes with (C++Std [intro.multithread]) the invocations of `f`. <br/> <br/> `bulk_execute` shall not propagate any exception thrown by `DECAY_COPY( std::forward<F>(f))(i, s)` or any other function submitted to the executor. [*Note:* The treatment of exceptions thrown by bulk one-way submitted functions and the forward progress guarantee of execution agents created by one-way execution functions is specific to the concrete executor type. *--end note.*] |
```
chriskohlhoff: This section still uses the phrasing "function submitted to the executor", "submitted functions", etc. Should it be changed to the new "callable object" formulation too?

wording.md
```diff
@@ -1603,7 +1617,7 @@ The `static_thread_pool` member functions `executor`, `attach`, `wait`, and
 do not introduce data races as a result of concurrent calls to those
 functions from different threads of execution.
 
-A `static_thread_pool`'s threads execute execution agents created via its associated executors with forward progress guarantee delegation. [*Note:* Forward progress is delegated to an execution agent for its lifetime. Because `static_thread_pool` guarantees only parallel forward progress to execution agents created via its executors, forward progress delegation does not apply to execution agents which have not yet started executing their first execution step. *--end note*]
+A `static_thread_pool`'s threads run execution agents with forward progress guarantee delegation. [*Note:* Forward progress is delegated to an execution agent for its lifetime. Because `static_thread_pool` guarantees only parallel forward progress to running execution agents; _i.e._, execution agents which have run the first step of the function object. *--end note*]
```
chriskohlhoff: I think the word "Because" should now be removed.

wording.md
hcedwar: Yes

wording.md
hcedwar: Yes

wording.md
nat-goodspeed: Is it significant that in the old wording, `sf()` is explicitly invoked on an unspecified execution agent, whereas in the new, the execution agent is simply unspecified? I don't know if there is a formal difference.

wording.md
nat-goodspeed: If I understand correctly, the intent of this change is to avoid stating that there will be an execution agent per work item; rather that the width of the thread pool determines the number of execution agents.

I like that formulation, though to be perfectly honest, I'm not entirely sure it's aligned with the definition of parallel forward progress guarantees. Paraphrasing, I believe the definition states that a parallel thread of execution runs concurrently with other parallel threads of execution once it has taken the first step; but that it does not guarantee to take the first step. That appears to require conceptualizing a thread of execution per work item. Please tell me I'm wrong about that?

This may be a situation in which we can benefit from having distinct definitions for "execution agent" and "thread of execution." Might we reasonably claim that there is a *thread of execution* per work item, but only as many *execution agents* as the width of the thread pool? In that case, there will come a point when the thread pool implementation (the execution context) associates an available execution agent with a work item (or perhaps "with the thread of execution associated with the work item"), thereby causing the work item's thread of execution to take the first step.

If by "forward progress guarantee delegation" we are referring to "boost-blocking," I don't think we can discuss that solely internally to the thread pool. The last-ditch bottom-line forward progress guarantee for a thread pool full of parallel execution agents is when some concurrent execution agent *external to the thread pool* demands a result -- in fact, blocks waiting for that result. At that point, it can lend its concurrent execution agent to the thread pool to ensure that at least one thread of execution takes at least one step.

But do we even need to discuss boost-blocking here? Isn't that covered in the section on forward progress guarantees? Couldn't the Note simply state that the `static_thread_pool` guarantees parallel forward progress, and leave it at that?

If "forward progress guarantee delegation" does not here mean boost-blocking, then we need to be clearer about what it does mean.

nat-goodspeed: n.b. I agree with Chris: the word "Because" doesn't fit in the current phrasing of the Note.

hcedwar:
@chriskohlhoff , @AerialMantis , @jaredhoberock 
Revised PR #402 to explicitly call out execution agent lifetime (front matter) and for operational semantics tables:
- remove requirement to immediately create execution agents
- clean up wording to match sequencing of actions; e.g., evaluate DECAY_COPY on the function before calling the resulting callable
- introduce a symbol for the callable to reduce repeated text

wording.md
```diff
@@ -18,6 +18,21 @@ inline namespace executors_v1 {
...
+
+As defined in [thread.req.lockable.general]:
+
+> "An *execution agent* is an entity such as a thread that may perform work in parallel with other execution agents. [*Note:* Implementations or users may introduce other kinds of agents such as processes or thread-pool tasks. *--end note*] The calling agent is determined by context; e.g., the calling thread that contains the call, and so on." 
+
+An execution agent supports execution by invoking a callable within an *execution context* such as as the calling thread or thread-pool.  An *executor* submits a callable to an execution context to be invoked by an execution agent within an associated execution context. [*Note:* Invocation of the callable may be inlined such as when the execution context is the calling thread, or may be scheduled such as when the execution context is a thread-pool with task scheduler. *--end note*] An executor may submit a callable with *execution properties* that specify how the submission and invocation of the callable interacts with the submitting thread and execution context, including forward progress guarantees [intro.progress].
```
jaredhoberock: Is a thread really an execution context? I thought we had been classifying threads as execution resources.

wording.md
```diff
@@ -466,13 +493,13 @@ The directionality properties conform to the following specification:
     {
       static constexpr bool is_requirable = true;
       static constexpr bool is_preferable = false;
-
+
```
jaredhoberock: Looks like there's some stray whitespace introduced here and below.

wording.md
```diff
@@ -700,9 +727,9 @@ The `blocking_t` property describes what guarantees executors provide about the
 
 | Nested Property Type | Nested Property Object Name | Requirements |
 |--------------------------|------------------------|--------------|
-| `blocking_t::possibly_t` | `blocking_t::possibly` | A call to an executor's execution function may block pending completion of one or more of the execution agents created by that execution function. |
-| `blocking_t::always_t` | `blocking_t::always` | A call to an executor's execution function shall block until completion of all execution agents created by that execution function. |
-| `blocking_t::never_t` | `blocking_t::never` | A call to an executor's execution function shall not block pending completion of the execution agents created by that execution function. |
+| `blocking_t::possibly_t` | `blocking_t::possibly` | A call to an executor's execution function may block pending completion of all invocations of the submitted callable object. |
```
jaredhoberock: Looks like we've changed "one or more" to "all" in some of these requirements. Do we want to make that change?

hcedwar:
Rebased, updated as per telecon review, and tightened up wording (see commit).

hcedwar:
@chriskohlhoff and @jaredhoberock : ping

jaredhoberock:
Merged.

PR #389, opened 2018-05-04
Allocator property fixes

chriskohlhoff:
Fixes for various bugs in the `allocator_t` specification, which I found when implementing this in asio.

PR #424, opened 2018-10-01
Slim down the proposal to target C++20

chriskohlhoff:


wording.md
```diff
@@ -639,23 +484,6 @@ polymorphic_executor_type(polymorphic_executor_type&& e) noexcept;
 
 *Effects:* If `!e`, `*this` has no target; otherwise, moves `e.target()` or move-constructs the target of `e` into the target of `*this`, leaving `e` in a valid state with an unspecified value.
 
-```
```
AerialMantis: Do we need to remove the narrowing constructor for oneway executors as well?

dhollman:
Should we be targeting C++20 with all of the properties currently in the paper? I don't feel all that comfortable getting behind some of the newer properties like `bulk_guarantee_t` and `outstanding_work_t` that I haven't seen any implementation experience around (correct me if I'm wrong about this!).

chriskohlhoff:
> Should we be targeting C++20 with all of the properties currently in the paper? I don't feel all that comfortable getting behind some of the newer properties like `bulk_guarantee_t` and `outstanding_work_t` that I haven't seen any implementation experience around (correct me if I'm wrong about this!).

The `outstanding_work_t` property is used by the networking TS so it will either end up here or there, but including it here means that `static_thread_pool` can support it too. It only has about 15 years implementation experience ;-) -- well, at least that since it was hardly an original idea when i first used it back then either.

I can't speak to `bulk_guarantee_t`, perhaps Jared can comment on its use.

jaredhoberock:
> I can't speak to bulk_guarantee_t, perhaps Jared can comment on its use.

Execution policies use those properties. See P1019.

dhollman:
Let's move this to an issue. I've opened issue #425 on this.

dhollman:
@jaredhoberock agreed, but is P1019 likely to make it into IS20? If not, it seems reasonable to put `bulk_guarantee_t` in a TS.

jaredhoberock:
I have a number for the separate paper: P1244.

chriskohlhoff:
I just noticed we never merged the bug fixes in #389. (My fault I think as I forgot to address Jared's comment.) Should we?

jaredhoberock:
Merged.

wording.md
AerialMantis: Sorry I'm late in replying to this. I was just wondering because it appeared as though the narrowing constructor wording was removed entirely, but I see now that they were moved to the oneway and bulk oneway polymorphic wrapper classes.

PR #431, opened 2019-02-05
Use Lee's proposed stronger wording for mapping_t::thread_t

### LINKED ISSUE #430: execution::mapping.thread includes things like fibers and SIMD lanes
`execution::mapping.thread`'s requirements say this:

> Execution agents are mapped onto threads of execution.

This language is broader than we would like and would seem to include weak things like fibers and SIMD lanes. We should clarify what we mean here.

Additionally, `execution::new_thread` may be unnecessary for C++20, since it was added to satisfy `std::async`'s requirements, and we're not currently planning to enhance `std::async` for C++20.

jaredhoberock:
This WIP uses Lee's proposed wording for `mapping_t::thread_t` and also eliminates mapping_t::thread_t`.

This PR is intended to resolve #430.

### DIFF
--- a/wording.md
+++ b/wording.md
@@ -925,16 +925,12 @@ The `mapping_t` property describes what guarantees executors provide about the m
 
 | Nested Property Type| Nested Property Object Name | Requirements |
 |-------------------------|---------------------------------|--------------|
-| `mapping_t::thread_t` | `mapping::thread` | Execution agents are mapped onto threads of execution. |
-| `mapping_t::new_thread_t` | `mapping::new_thread` | Each execution agent is mapped onto a new thread of execution. |
+| `mapping_t::thread_t` | `mapping::thread` | An agent A has exclusive ownership of a `std::thread` for the duration of A's lifetime and any agents to which A delegates forward progress. |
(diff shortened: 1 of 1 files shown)


jaredhoberock:
I've updated the PR to replace the use of `std::thread` with "thread of execution". I've also eliminated the note regarding TLS. Since not all threads of execution may have TLS, the note is no longer correct.

wording.md
```diff
@@ -925,17 +925,9 @@ The `mapping_t` property describes what guarantees executors provide about the m
 
 | Nested Property Type| Nested Property Object Name | Requirements |
 |-------------------------|---------------------------------|--------------|
-| `mapping_t::thread_t` | `mapping::thread` | Execution agents are mapped onto threads of execution. |
-| `mapping_t::new_thread_t` | `mapping::new_thread` | Each execution agent is mapped onto a new thread of execution. |
+| `mapping_t::thread_t` | `mapping::thread` | An agent A has exclusive ownership of a thread of execution for the duration of A's lifetime and any agents to which A delegates forward progress. |
 | `mapping_t::other_t` | `mapping::other` | Mapping of each execution agent is implementation-defined. |
```
griwes: The modified rows of the above table should be unified with the proposed changes in #436.

PR #434, opened 2019-02-11
All-or-nothing guarantee for bulk_execute

### LINKED ISSUE #432: Consider all-or-nothing guarantee for bulk_execute
`bulk_execute`'s operational semantics include this:

>Shall not propagate any exception thrown by cf or any other function submitted to the executor. [Note: The treatment of exceptions thrown by bulk one-way submitted functions and the forward progress guarantee of the associated execution agent(s) are implementation defined. --end note.]

This language seems to permit the executor to fail via exception in the middle of the submission, having created a partial group of execution agents. The problem is that it's difficult to see how a client could handle that exception without a universal way of communicating which agents were successfully created. 

One option would be to ban `bulk_execute` from creating agents if an exception is thrown. Would that be too restrictive?

jaredhoberock:
This is intended to resolve #432


jaredhoberock:
Say "complete group of *n* execution agents"

PR #440, opened 2019-09-30
WIP: Sender integration

jaredhoberock:


jaredhoberock:
This initial PR is very rough. Please see the front matter's TODO list for what still needs to be introduced into the wording.

wording.md
```diff
@@ -1603,56 +1509,42 @@ bool operator!=(const C& a, const C& b) noexcept;
 
 *Returns:* `!(a == b)`.
 
-#### `static_thread_pool` executor types with the `execution::oneway` property
+#### `static_thread_pool` executor execution functions
```
kirkshoop: won't let me comment on unchanged text - scheduler and schedule should be included from P1660 and the static_thread_pool should have a using for `scheduler_type` and `scheduler_type scheduler() noexcept;` method to match the `executor_type executor() noexcept;` method

wording.md
```diff
@@ -1603,56 +1509,42 @@ bool operator!=(const C& a, const C& b) noexcept;
...
-  public:
-    template<class Function, class SharedFactory>
-      void bulk_execute(Function&& f, size_t n, SharedFactory&& sf) const;
-};
-```
-
-`C` is a type satisfying the `BulkOneWayExecutor` requirements.
+according to the the properties established for `*this`. If the submitted
+function `f` exits via an exception, the `static_thread_pool` invokes
+`std::terminate()`.
```
kirkshoop: the scheduler version can do better than `terminate();`

wording.md
```diff
@@ -183,6 +225,72 @@ This sub-clause contains a template that may be used to query the applicability
...
+
+- Otherwise, `bulk_execute(E, F, S)`, if that expression is valid, with overload resolution performed in a context that includes the declaration
+
+      template<class E, class F, class S>
+        void bulk_execute(E, F, S) = delete;
+
+    and that does not include a declaration of `execution::bulk_execute`. If the function selected by overload resolution does not execute `S` invocations of the function object `F` on the executor `E` in bulk, and the result of that function does not model `sender<void>`, the program is ill-formed with no diagnostic required.
+
+- Otherwise, if the types `F` and `executor_index_t<remove_cvref_t<E>>` model `invocable`, then `execution::tbd_concrete_implementations::bulk_execute(E, F, S)`.
+
```
kirkshoop: `tbd_concrete_implementations` came from Lee's slide - I am not sure what it means here. The execute description above does not mention this namespace.

wording_front_matter.md
```diff
@@ -19,11 +19,11 @@ Authors:            Jared Hoberock, jhoberock@nvidia.com
...
                     Lee Howes, lwh@fb.com
 
-                    Kirk Shoop, XXX get Kirk's email
+                    Kirk Shoop, kirkshoop@fb.com
 
-                    Lewis Baker, XXX get Lewis' email
+                    Lewis Baker, lbaker@fb.com
 
-                    Eric Niebler, XXX get Eric's email
+                    Eric Niebler, eniebler@fb.com
```
kirkshoop: will the abstract still be accurate after P1660 is applied?

"Abstract:           This paper proposes a programming model for executors, which are modular components for creating execution. The design of this proposal is described in paper [P0761](https://wg21.link/P0761)."

wording.md
```diff
@@ -183,6 +225,72 @@ This sub-clause contains a template that may be used to query the applicability
 |----------------------------|---------------------|----------------|
 | `template<class T>` <br/>`struct is_executor` | The expression `P::is_executor` is a well-formed constant expression with a value of `true`. | `T` is a complete type. |
 
+### Customization points
+
+#### `execution::execute`
+
+The name `execution::execute` denotes a customization point object. The expression `execution::execute(E, F)` for some subexpressions `E` and `F` is expression-equivalent to:
+
+- `E.execute(F)`, if that expression is valid. If the function selected does not execute the function object `F` on the executor `E`, the program is ill-formed with no diagnostic required.
```
AerialMantis: I'm not sure if I understand the "the program is ill-formed with no diagnostic required", is the intention to allow the executor to not handle the error if it fails to invoke `F` on `E`?

wording.md
```diff
@@ -183,6 +225,72 @@ This sub-clause contains a template that may be used to query the applicability
...
+- `S.submit(C)`, if that expression is valid. If the function selected does not submit the callback object `C` via the sender `S`, the program is ill-formed with no diagnostic required.
+
+- Otherwise, `submit(S, C)`, if that expression is valid, with overload resolution performed in a context that includes the declaration
+
+        template<class S, class C>
+          void submit(S, C) = delete;
+
+    and that does not include a declaration of `execution::submit`. If the function selected by overload resolution does not submit the callback object `C` via the sender `S`, the program is ill-formed with no diagnostic required.
+
+- Otherwise, if the type `C` models `Callback`, then let `F = [C]{ C.value(); }`. If `S` and `F` model `executor`, then `execution::execute(S, F)`.
```
AerialMantis: We mention `value` here but it's not defined as part of the `callback` concept.

AerialMantis: I'm a bit confused by the wording here - "If `S` and `F` model `executor`", is this a typo?

AerialMantis: As currently define this adaptation could result in a call to `execution::submit` with a function who's type models `callback` where the `error` and `done` channels are not honoured, is this expected?

wording.md
```diff
@@ -183,6 +225,72 @@ This sub-clause contains a template that may be used to query the applicability
...
+- Otherwise, `execution::submit(S, C)` is ill-formed.
+
+[*Editorial note:* We should probably define what "submit the callback object `C` via the sender `S`" means more carefully. *--end editorial note*]
+
+[*Editorial note:* This specification is adapted from `ranges::iter_swap`. *--end editorial note*]
+
+#### `execution::bulk_execute`
+
+The name `execution::bulk_execute` denotes a customization point object. If `is_convertible_v<S, executor_shape_t<remove_cvref_t<E>>>` is true, then the expression `execution::bulk_execute(E, F, S)` for some subexpressions `E`, `F`, and `S` is expression-equivalent to:
+
```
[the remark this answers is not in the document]
AerialMantis: I think `size_t` here is fine for now as well.

wording_front_matter.md
```diff
@@ -59,9 +59,23 @@ Abstract:           This paper proposes a programming model for executors, which
 
 As directed by SG1 at the 2019-XX Cologne meeting, we have implemented the following changes suggested by P1658 and P1660:
 
-* XXX TODO
-* XXX TODO
-* XXX TODO
+* Eliminated interface-changing properties `oneway_t` and `bulk_oneway_t`.
+* Introduced `Executor` concept.
```
AerialMantis: `Executor` -> `executor`.

wording.md
```diff
@@ -183,6 +225,72 @@ This sub-clause contains a template that may be used to query the applicability
...
+- `E.execute(F)`, if that expression is valid. If the function selected does not execute the function object `F` on the executor `E`, the program is ill-formed with no diagnostic required.
+
+- Otherwise, `execute(E, F)`, if that expression is valid, with overload resolution performed in a context that includes the declaration
+
+        template<class E, class F>
+          void execute(E, F) = delete;
+
+    and that does not include a declaration of `execution::execute`. If the function selected by overload resolution does not execute the function object `F` on the executor `E`, the program is ill-formed with no diagnostic required.
+
+- Otherwise, if the type `F` models `invocable`, then `execution::submit(E, execution::make_callback_from(F))`.
```
AerialMantis: `execution::make_callback_from` is used here but it's not defined.

wording.md
```diff
@@ -183,6 +225,72 @@ This sub-clause contains a template that may be used to query the applicability
...
+
+- Otherwise, `execution::execute(E, F)` is ill-formed.
+
+[*Editorial note:* We should probably define what "execute the function object `F` on the executor `E`" means more carefully. *--end editorial note*]
+
+[*Editorial note:* This specification is adapted from `ranges::iter_swap`. *--end editorial note*]
+
+#### `execution::submit`
```
AerialMantis: There is no wording for `execution::submit(executor, callback)`, though it is used in the adaptation for `execution::execute(executor, invocable)`. Unless we assume that `executor` is a refinement of `sender`, though I believe we decided to define `sender` and `executor` as separate concepts.

wording.md
```diff
@@ -1603,56 +1509,42 @@ bool operator!=(const C& a, const C& b) noexcept;
...
 
 ```
 class C
 {
   public:
     template<class Function>
       void execute(Function&& f) const;
+
+    template<class Function, class SharedFactory>
+      void bulk_execute(Function&& f, size_t n, SharedFactory&& sf) const;
```
AerialMantis: Should we remove the `SharedFactory` parameter from `bulk_execute`?

wording.md
AerialMantis: I am also unsure exactly what this means, is the intention that the standard will define a concrete implementation of `bulk_execute` as a fall-back adaptation, such as a for loop over `execution::execute`?

wording.md
```diff
@@ -121,55 +129,89 @@ namespace execution {
...
+
+None of these operations, nor an executor type's `execute` or `submit` functions, or associated query functions shall introduce data races as a result of concurrent invocations of those functions from different threads.
 
-A type `X` satisfies the `BulkOneWayExecutor` requirements if it satisfies the general requirements on executors, as well as the requirements in the Table below.
+For any two (possibly const) values `x1` and `x2` of some executor type `X`, `x1 == x2` shall return `true` only if `x1.query(p) == x2.query(p)` for every property `p` where both `x1.query(p)` and `x2.query(p)` are well-formed and result in a non-void type that is `EqualityComparable` (C++Std [equalitycomparable]). [*Note:* The above requirements imply that `x1 == x2` returns `true` if `x1` and `x2` can be interchanged with identical effects. An executor may conceptually contain additional properties which are not exposed by a named property type that can be observed via `execution::query`; in this case, it is up to the concrete executor implementation to decide if these properties affect equality. Returning `false` does not necessarily imply that the effects are not identical. *--end note*]
```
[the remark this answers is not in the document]
kirkshoop: The equality requirement is in the concept now. If the concept clause needs to be repeated in the normative prose then equality is not sufficient.

wording.md
```diff
@@ -121,55 +129,89 @@ namespace execution {
...
-| Expression | Return Type | Operational semantics |
-|------------|-------------|---------------------- |
-| `x.execute(f)` | `void` | Evaluates `DECAY_COPY(std::forward<F>(f))` on the calling thread to create `cf` that will be invoked at most once by an execution agent. <br/> May block pending completion of this invocation. <br/> Synchronizes with [intro.multithread] the invocation of `f`. <br/>Shall not propagate any exception thrown by the function object or any other function submitted to the executor. [*Note:* The treatment of exceptions thrown by one-way submitted functions is described by the `execution::oneway_exception_handler` property. The forward progress guarantee of the associated execution agent(s) is implementation defined. *--end note.*] |
+  (
+    requires(E&& e, F&& f) {
+      ((E&&)e).execute((C&&)c);
+    }
 
-### `BulkOneWayExecutor` requirements
+    ||
```
ericniebler: For many reasons, I don't think we want the `executor` concept to be a disjunction of unrelated capabilities. I suggest an `executor` is something that can be passed to the `execute` CPO, and `sender`/`scheduler` are defined separately.

wording.md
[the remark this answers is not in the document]
ericniebler: The fact that `x1` and `x2` are _equal_ (in the sense of [\[concepts.equality\]](http://eel.is/c++draft/concepts.equality)) does not necessarily imply that `x1.foo(bar)` and `x2.foo(bar)` are _equal_, so I don't think this paragraph can be removed. Not yet, at least.

I agree with Kirk's earlier statement that it would make P0443 easier to review in LEWG if we separated the executor properties out into a different proposal.

wording.md
ericniebler: This formulation of `executor` also leaves us in the awkward situation where `executor<E,F>` could be `true`, but `executor<E>` is `false`. We spitballed something like the following in Cologne:

```c++
template<class E, class F>
concept executor-impl =
    invocable<F> &&
    equality_comparable<E> &&
    requires (E&& e, F&& f) {
      execute((E&&) e, (F&&) f);
    };

template<class E>
concept executor = executor-impl<E, void(*)()>;

template<class E, class F>
concept executor_of =
  executor<E> &&
  executor-impl<E, F>;
```

This would require _all_ executors to have the ability to at least execute a `void(*)()`. Then we could drop the `is_executor_v` trait because we could use the concept to test whether a type models `executor` or not.

wording.md
```diff
@@ -19,24 +19,37 @@ For the intent of this library and extensions to this library, the *lifetime of
...
+    concept callback_signal = see-below;
+
+  template<class T, class... an>
+    concept callback = see-below;
+
+  template<class S, class C>
+    concept sender_to = see-below;
+
   // Indication of executor property applicability
   template<class T> struct is_executor;
```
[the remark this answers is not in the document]
jaredhoberock: Me too. It wasn't immediately clear to me that the current definition of `executor` (which is a binary type concept) is a complete replacement for `is_executor` (a unary type trait), so I left it in to be conservative.

wording.md
```diff
@@ -121,55 +129,89 @@ namespace execution {
...
+XXX TODO The `callback_signal` concept...
 
-None of these concepts' operations, nor an executor type's swap operations, shall exit via an exception.
+```
+template<class T, class E = exception_ptr>
+concept callback_signal =
+  requires(T&& t, E&& e) {
+    { ((T&&) t).done()} noexcept;
+    { ((T&&) t).error(((E&&) e)) } noexcept;
+  };
```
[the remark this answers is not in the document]
jaredhoberock: OK, I'll implement this suggestion before the call.

wording.md
jaredhoberock: @AerialMantis, yes, I think that was what Lee was getting at. It seemed unconventional to me, so I think we should discuss on the call (unless @LeeHowes wants to chime in here).

wording.md
```diff
@@ -121,55 +129,89 @@ namespace execution {
...
+| `e.submit(f)` | `void` | Evaluates `DECAY_COPY(std::forward<F>(f))` on the calling thread to create `cf` that will be invoked at most once by an execution agent. <br/> May block pending completion of this invocation. <br/> Synchronizes with [intro.multithread] the invocation of `f`. <br/>Shall not propagate any exception thrown by the function object or any other function submitted to the executor. [*Note:* The treatment of exceptions thrown by one-way submitted functions is described by the `execution::oneway_exception_handler` property. The forward progress guarantee of the associated execution agent(s) is implementation defined. *--end note.*] |
```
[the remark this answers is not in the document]
jaredhoberock: Yes, somewhere in the paper the callback contract needs to be described. Maybe in the `callback` concept definition we can add a paragraph defining the `callback` contract so that other parts of the paper can simply refer to it without repetition.

wording.md
jaredhoberock: Right. I wasn't sure if @LeeHowes wanted `make_callback_from` to be an exposition-only function or not.

wording.md
LeeHowes: Isn't this inevitable? I don't care whether we define it by calling out to a known standard library function, or describe it as being the default implementation inside the CPO, but I think all algorithms need some sort of default implementation.

For simple algorithms like transform the default standard library version is trivially what we would use for most network code. It would wrap the function in a callback, and pass that to submit on the incoming sender when submit is called on the returned sender.

For a complex algorithm like bulk_execute (or more so, sort) it would fall back to whatever Microsoft or others think is a sensible default. That is, Billy can build whatever extension he likes based, presumably, on what is in std::parallel_for, std::sort etc.

I strongly believe that if we do not want every executor author to have to customise every algorithm, we need defaults.

wording.md
```diff
@@ -121,55 +129,89 @@ namespace execution {
...
+| `e.execute(f)` | `void` | Evaluates `DECAY_COPY(std::forward<F>(f))` on the calling thread to create `cf` that will be invoked at most once by an execution agent. <br/> May block pending completion of this invocation. <br/> Synchronizes with [intro.multithread] the invocation of `f`. <br/>Shall not propagate any exception thrown by the function object or any other function submitted to the executor. [*Note:* The treatment of exceptions thrown by one-way submitted functions is described by the `execution::oneway_exception_handler` property. The forward progress guarantee of the associated execution agent(s) is implementation defined. *--end note.*] |
```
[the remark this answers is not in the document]
LeeHowes: I think that the cleaner solution is to define Callback so it is not also an Invocable, that way you have a clean compile-time failure instead of a contract failure. execute and submit being mutually callable with a default and wrapping operation covers the interop node without subsumption of the actual callback/invocable concepts.

wording.md
LeeHowes: I used it as a placeholder.

I think the cleanest way to define the default version of execute though is to defined make_callback_from and use it. Optimally, make_callback_from should be fairly general though so that the default for transform et al in future can be defined in terms of it. 

Kirk probably has a nice definition but I imagine it would be something like:

Callback make_callback_from(Invocable f, Callback downstream) {
  struct CB {
    Optional<Invocable> f_;
    void value() noexcept {
      try {
        auto v = f();
        value(downstream, std::move(v));
      } catch(...) {
         auto ep = std::current_exception();
         error(downstream, std::move(ep));
      }
    }
    void error(E err) noexcept {      
      error(downstream, std::move(err));
    }
    void done() noexcept {
       f_ = Optional<Invocable>{};
       done(downstream);
    }
  };
}

and default for execute would be:
execution::submit(E, execution::make_callback_from(F, null_callback));

wording.md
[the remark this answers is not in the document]
jaredhoberock: Do these customization points have defaults?

wording.md
AerialMantis: One concern I have about the minimum requirement for `executor` being that it can call a `void(*)()` that this may be problematic for some executors to guarantee. For example, a SYCL-based executor would only be able to support `execute(T)` were `T` is a non-type-erased function object or lambda expression. Is there an alternative requirement we could use here that wouldn't enforce supporting invoking a function pointer?

LeeHowes: I think that you should treat execute(f) and submit(c) as being *host* operations if called, rather than try to work around limitations in this way. Use the customised algorithms where you want to restrict. Otherwise you risk being in a situation where I pass an interop callback to submit, you wrap it into execute and the system surprisingly fails because it turns out to be host code you can't run on the device. These default interfaces are the fallback safe interop interfaces, not the ones we use for offload workloads.

Having said that, if this constraint leads to us also wondering how to define bulk_execute's invocable parameter, then I think this is a serious question to consider.

wording.md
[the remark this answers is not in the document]
AerialMantis: I agree, this would be my preferred approach as well, I realised just after the last call ended that we didn't discuss whether to maintain the subsumption hierarchy of `invocable` and `callback`, but I feel that separating them makes the most sense with the direction we're taking.

With the change that @ericniebler proposed above of separating the concepts for `executor` and `sender` as requiring `execute` and `submit` respectively, I believe this meansa we now have:

The `executor` concept requiring `tbd::execute(executor, invocable)` and the `sender` concept requiring `tbd::submit(sender, callback)`, and no subsumption between them, does that sound right?

Another question this raises for me is, does this mean we should re-introduce the `value` channel on the `callback` concept as well? Because the current PR follows P1660 which uses the function call operation in order to subsum `invocable`.

wording.md
AerialMantis: This makes sense to me, though I think we need to specify the requirements in such a way to allow `f` to be `noexcept` and allow implementations to not introduce error handling in the adaption.

wording.md
```diff
@@ -121,55 +129,89 @@ namespace execution {
...
+  copy_constructible<remove_cvref_t<E>> &&
+  equality_comparable<remove_cvref_t<E>> &&
 
-| Expression | Return Type | Operational semantics |
-|------------|-------------|---------------------- |
-| `x.execute(f)` | `void` | Evaluates `DECAY_COPY(std::forward<F>(f))` on the calling thread to create `cf` that will be invoked at most once by an execution agent. <br/> May block pending completion of this invocation. <br/> Synchronizes with [intro.multithread] the invocation of `f`. <br/>Shall not propagate any exception thrown by the function object or any other function submitted to the executor. [*Note:* The treatment of exceptions thrown by one-way submitted functions is described by the `execution::oneway_exception_handler` property. The forward progress guarantee of the associated execution agent(s) is implementation defined. *--end note.*] |
+  (
+    requires(E&& e, F&& f) {
+      ((E&&)e).execute((C&&)c);
+    }
```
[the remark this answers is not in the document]
jaredhoberock: Done.

wording.md
```diff
@@ -121,55 +129,89 @@ namespace execution {
...
 
-### `BulkOneWayExecutor` requirements
+    ||
 
-The `BulkOneWayExecutor` requirements specify requirements for executors which submit a function object to be invoked multiple times without a channel for awaiting the completion of the submitted function object invocations and obtaining their result. [*Note:* That is, the executor provides fire-and-forget semantics. *--end note*]
+    requires(E&&e, F&& f) {
+      ((E&&)e).submit((F&&)f);
+    }
+  );
```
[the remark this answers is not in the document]
jaredhoberock: Done.

wording.md
```diff
@@ -1603,56 +1509,42 @@ bool operator!=(const C& a, const C& b) noexcept;
...
-```
-
-`C` is a type satisfying the `BulkOneWayExecutor` requirements.
+according to the the properties established for `*this`. If the submitted
+function `f` exits via an exception, the `static_thread_pool` invokes
+`std::terminate()`.
 
 ```
 template<class Function, class SharedFactory>
   void bulk_execute(Function&& f, size_t n, SharedFactory&& sf) const;
```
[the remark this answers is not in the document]
jaredhoberock: Done

wording.md
jaredhoberock: I think we all agree that a default implementation for `execution::bulk_execute` is necessary. What was unclear was whether a named function corresponding to that default implementation is necessary. I think I'd prefer avoiding introducing another named function and just specify what `execution::bulk_execute`'s default is (i.e., execute in a sequential `for` loop).

wording.md
```diff
@@ -121,55 +129,89 @@ namespace execution {
...
+  invocable<F> &&
+  copy_constructible<remove_cvref_t<E>> &&
+  equality_comparable<remove_cvref_t<E>> &&
 
-| Expression | Return Type | Operational semantics |
-|------------|-------------|---------------------- |
-| `x.execute(f)` | `void` | Evaluates `DECAY_COPY(std::forward<F>(f))` on the calling thread to create `cf` that will be invoked at most once by an execution agent. <br/> May block pending completion of this invocation. <br/> Synchronizes with [intro.multithread] the invocation of `f`. <br/>Shall not propagate any exception thrown by the function object or any other function submitted to the executor. [*Note:* The treatment of exceptions thrown by one-way submitted functions is described by the `execution::oneway_exception_handler` property. The forward progress guarantee of the associated execution agent(s) is implementation defined. *--end note.*] |
+  (
+    requires(E&& e, F&& f) {
+      ((E&&)e).execute((C&&)c);
```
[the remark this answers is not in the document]
jaredhoberock: Done

wording.md
```diff
@@ -121,55 +129,89 @@ namespace execution {
...
+    }
 
-### `BulkOneWayExecutor` requirements
+    ||
 
-The `BulkOneWayExecutor` requirements specify requirements for executors which submit a function object to be invoked multiple times without a channel for awaiting the completion of the submitted function object invocations and obtaining their result. [*Note:* That is, the executor provides fire-and-forget semantics. *--end note*]
+    requires(E&&e, F&& f) {
+      ((E&&)e).submit((F&&)f);
```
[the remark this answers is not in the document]
jaredhoberock: Good question. Your suggestion sounds correct, but I'm not sure how to enforce that in the concept definition.

wording.md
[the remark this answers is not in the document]
kirkshoop: the defaults look for the member methods.

wording.md
jaredhoberock: I think what that language means is that if `E.execute(F)` compiles but `E` doesn't actually execute `F` (i.e., it reneges on the `execute` contract), then the program is bogus. "No diagnostic required" means e.g. the compiler doesn't have to complain about this case (because it couldn't detect it anyway).

wording.md
jaredhoberock: > We mention value here but it's not defined as part of the callback concept.

Right, I think we need to decide whether `callback` should use the call operator or `value` for the value channel. The comments in this review suggest that we may prefer a named function `value`.

> I'm a bit confused by the wording here - "If S and F model executor", is this a typo?

`executor<E,F>` has two type parameters, so two types participate in modeling. Not sure if that language is the best way to say that, or if there are already examples of such language in the C++23 draft.

> As currently define this adaptation could result in a call to execution::submit with a function who's type models callback where the error and done channels are not honoured, is this expected?

We should think through this case.

wording.md
kirkshoop: we can add the function and describe its behaviour while leaving the name unspecified, yes?

wording.md
jaredhoberock: Done

wording.md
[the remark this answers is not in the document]
jaredhoberock: If the member methods do not exist, is there a fallback for these CPOs?

wording.md
kirkshoop: I have this code in [quick-bench](http://quick-bench.com/Dhr1upUkfA4bJdKt4-b7mrmXf1U)

```cpp
template<class Callback>
struct invocable_from_fn {
  Callback cb_;
  bool need_done_;
  ~invocable_from_fn() {
    if (need_done_) {
      cb_.done();
    }
  }
  template<class Callback_>
  explicit invocable_from_fn(Callback_&& cb) noexcept try : 
    cb_(std::forward<Callback_>(cb)),
    need_done_(true) {
  } catch(...) {
    cb.error(std::current_exception());
  }
  invocable_from_fn(const invocable_from_fn&) = delete;
  invocable_from_fn& operator=(const invocable_from_fn&) = delete;
  invocable_from_fn(invocable_from_fn&& o) noexcept try : 
    cb_(std::move(o.cb_)), 
    need_done_(false) {
    need_done_ = o.need_done_;
    o.need_done_ = false;
  } catch (...) {
    o.cb_.error(std::current_exception());
    o.need_done_ = false;
  }
  invocable_from_fn& operator=(invocable_from_fn&& o) noexcept {
    if (need_done_) {
      cb_.done();
    }
    try {
      cb_ = std::move(o.cb_);
      need_done_ = o.need_done_;
      o.need_done_ = false;
    } catch (...) {
      o.cb_.error(std::current_exception());
      o.need_done_ = false;
      need_done_ = false;
    }
    return *this;
  }
  template<class... An>
  void operator()(An&&... an) noexcept(std::is_nothrow_invocable_v<Callback&, An...>) {
    // move errors are being suppressed to prevent terminate();
    // use need_done_ to noop the thread after a failed move
    if (!need_done_) {return;}
    if constexpr (std::is_nothrow_invocable_v<Callback&, An...>) {
      std::invoke(cb_, std::forward<An>(an)...);
    } else {
      try {
        std::invoke(cb_, std::forward<An>(an)...);
      } catch (...) {
        cb_.error(std::current_exception());
      }
    }
    need_done_ = false;
  }
};
template<class Callback>
auto invocable_from(Callback&& cb) 
  -> invocable_from_fn<std::decay_t<Callback>> {
  return invocable_from_fn<std::decay_t<Callback>>{std::forward<Callback>(cb)};
}
```

kirkshoop: Lee's `make_callback_from()` is an implementation detail of the `Sender transform(Sender, Function)` algorithm. 

with `make_callback_from()`

```cpp
execution::submit(E, execution::make_callback_from(F, null_callback));
```

using `transform()` instead:

```cpp
execution::submit(tbd::transform(E, F), null_callback);
```

wording.md
LeeHowes: I actually think the for loop is over-constraining. The important thing is how it interacts with the Senders so we get predictable chaining. The execution of f we can tweak the wording from for_each:
"Effects: Submits a Callback c to inputSender.submit(c) such that when value() is called on c applies f to every value in the range [0, n). When all calls are complete, calls value() on the callback passed from submit() on the returned Sender. " or whatever works along those lines.

wording.md
LeeHowes: True, it is fundamentally transform, but we decided not to define transform at this time.

For very clean definitions of things on top of each other we should probably define one or other. This may be a good reason to pull scalar transform back into this, or we can debate later about removing make_callback and replacing it with transform once we add transform.

Or we can put prose in to define the actions of what I wrote in code.

wording.md
```diff
@@ -19,6 +19,22 @@ For the intent of this library and extensions to this library, the *lifetime of
...
 namespace execution {
 
+  // Customization points:
+
+  inline namespace unspecified{
+    inline constexpr unspecified value = unspecified;
+
+    inline constexpr unspecified done = unspecified;
+
+    inline constexpr unspecified value = unspecified;
```
kirkshoop: typo - perhaps value -> error?

wording.md
```diff
@@ -129,103 +138,64 @@ namespace execution {
...
-XXX TODO The `executor` concept...
+#### `execution::done`
 
-```
-template<class E, class F = void(*)()>
-concept executor =
-  invocable<F> &&
-  copy_constructible<remove_cvref_t<E>> &&
-  equality_comparable<remove_cvref_t<E>> &&
+The name `execution::value` denotes a customization point object. The expression `execution::done(C)` for some subexpression `C` is expression-equivalent to:
```
kirkshoop: also - how do we specify that V is a variadic pack?

wording.md
```diff
@@ -129,103 +138,64 @@ namespace execution {
...
+- Otherwise, `execution::value(C, V)` is ill-formed.
 
-```
-template<class S, class C>
-concept sender_to =
-  requires(S&& s, C&& c) {
-    submit((S&&) s, (C&&) c);
-  };
-```
+[*Editorial note:* We should probably define what "send the value `V` to the callback `C`'s value channel" means more carefully. *--end editorial note*]
```
kirkshoop: also - how do we specify that V is a variadic pack?

wording.md
kirkshoop: typo - value -> done

wording.md
```diff
@@ -129,103 +138,64 @@ namespace execution {
...
+The name `execution::value` denotes a customization point object. The expression `execution::done(C)` for some subexpression `C` is expression-equivalent to:
 
-  (
-    requires(E&& e, F&& f) {
-      ((E&&)e).execute((C&&)c);
-    }
+- `C.done()`, if that expression is valid. If the function selected does not signal the callback `C`'s done channel, the program is ill-formed with no diagnostic required.
 
-    ||
+- Otherwise, `done(C)`, if that expression is valid, with overload resolution performed in a context that includes the declaration
```
kirkshoop: done is required to be noexcept

wording.md
```diff
@@ -129,103 +138,64 @@ namespace execution {
...
-| `e.submit(f)` | `void` | Evaluates `DECAY_COPY(std::forward<F>(f))` on the calling thread to create `cf` that will be invoked at most once by an execution agent. <br/> May block pending completion of this invocation. <br/> Synchronizes with [intro.multithread] the invocation of `f`. <br/>Shall not propagate any exception thrown by the function object or any other function submitted to the executor. [*Note:* The treatment of exceptions thrown by one-way submitted functions is described by the `execution::oneway_exception_handler` property. The forward progress guarantee of the associated execution agent(s) is implementation defined. *--end note.*] |
+- Otherwise, `error(C, E)`, if that expression is valid, with overload resolution performed in a context that includes the declaration
```
kirkshoop: error must be noexcept

wording.md
```diff
@@ -77,6 +90,7 @@ namespace execution {
   // One-way exception handling property:
 
   // XXX bikeshed this name. "handler" may imply that the value of the property is some kind of function
+  // [*Editorial Note:* The presence of this property seems at-odds with the error-handling scheme adopted from P1660. Does this property need to be eliminated? *--end editorial note.*]
   struct oneway_exception_handler_t;
```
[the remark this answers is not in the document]
AerialMantis: I agree that a concept based approach may mean some properties should change or be removed, but I feel we should avoid changing these properties wholesale without having a discussion about it first.

wording.md
jaredhoberock: Done

wording.md
jaredhoberock: Done

wording.md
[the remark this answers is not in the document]
AerialMantis: This makes sense to me, that way we preserve the requirement of property equality, but we defer to the explicit wording for introducing a concept requirement for `query` to a future change.

As for separating properties from P0443 into a separate paper, I can see the argument for that, but I would prefer to avoid a substantial change list that until we have more time to discuss it properly and we get SG1's feedback on it.

wording.md
AerialMantis: I agree that we definitely need to define defaults and I would also prefer that we define these in the wording of the CPO as opposed to a named implementation. I think wording along those lines makes sense to me.

wording.md
LeeHowes: I think we should use the value channel and make sure that passing a Callback to execute is an error, and an Invocable to submit is an error. Clear compile-time checks will be much safer for code than the risk of accidentally not obeying the channels. Submit mapped to execute then has a clear implementation that is well-defined.

wording.md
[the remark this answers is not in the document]
AerialMantis: I was thinking we could have something like:
```
    callback<E> &&
    requires(E&&e, F&& f) {
      execution::submit((E&&)e,(F&&)f);
    }
```

Though if we were to go with @ericniebler's suggestion and separate the `executor` and `sender` concepts, I believe we could move out the part after the disjunction into the `sender` concept with something like:

```
template<class E, class C>
concept sender =
  callback<C> &&
  copy_constructible<remove_cvref_t<E>> &&
  equality_comparable<remove_cvref_t<E>> &&
  (
    requires(E&& e, C&& c) {
      execution::submit((E&&)e,(C&&)c);
    }
  );
  };
```

Though I am not too experienced with concepts so I someone may correct me on this.

wording.md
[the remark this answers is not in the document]
AerialMantis: Yeah that makes sense because we'll need to describe the same contract for both `sender::submit(callback)` and `tbd::submit(sender, callback)`.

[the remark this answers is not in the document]
AerialMantis: Perhaps we should do the same for `executor::execute(invocable)` and `tbd::execute(executor, invocable)`?

wording.md
AerialMantis: Okay, that makes sense, thanks!

wording.md
AerialMantis: > executor<E,F> has two type parameters, so two types participate in modeling. Not sure if that language is the best way to say that, or if there are already examples of such language in the C++23 draft.

Oh yeah, of course, I was forgetting that `executor` was defined in terms of two types.

AerialMantis: I like this solution, it avoids any ambiguities and means any adaptions between `execute` and `submit` or vice versa are well defined.

wording.md
AerialMantis: This makes sense to me, I'm not sure if we'd have time to discuss and review further algorithms for this revision, but I think this is a logical next step.

wording.md
AerialMantis: Okay, I see what you mean, we can have the `executor` concept require that it be capable of calling `execute(f)` where `f` is of type `void(*)())`, but then have later algorithms such as `tbd::bulk_execute(invocable, shape)` (assuming we define it) be customizable with restrictions.

So I could have an `opencl_executor` who has a `bulk_execute` that has further limitations of its `invocable` concept, but it would still provide `execute` which can execute function pointers in host code.

wording.md
LeeHowes: Scalar transform is so simple, and such an obvious thing to have in a pipeline, that it might be easier to explain than make_callback_from. Equally, describing make_callback_from in a few words of prose might be enough for Belfast.

wording.md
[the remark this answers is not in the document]
kirkshoop: I would not specify one for this revision. I would like to have a small, non-controversial paper for belfast

wording.md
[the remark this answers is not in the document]
kirkshoop: Yes, we will discuss in the call. I agree that no properties should be changed before Belfast. 

I believe that the items I listed above in 0443 increase risk due to controversy over their semantics (for example, should blocking.always mean that f is invoked inline or that the calling thread is blocked until f is invoked and synchronizes with the invoke of f, or should these be separated into something like will_inline() and will_wait()). Semantic controversies are independent of the controversy over the mechanics of how properties are defined and used. 

I believe that the cpo's, concepts and static_thread_pool are useful, without definitions for the items I listed above in the standard. 

I believe that the earlier 0443 goes in, the more time we have to work on other execution papers targeting '23.

I believe that removing algorithms and properties from 0443 for Belfast (just as ranges shipped a subset and coroutines shipped with little library support, and constexpr, and noexcept, etc..) will make it easier to vote the cpo's, concepts, and static_thread_pool into the working paper.

wording.md
```diff
@@ -121,55 +138,211 @@ namespace execution {
...
+- Otherwise, `execution::submit(S, C)` is ill-formed.
+
+[*Editorial note:* We should probably define what "submit the callback object `C` via the sender `S`" means more carefully. *--end editorial note*]
+
+[*Editorial note:* This specification is adapted from `ranges::iter_swap`. *--end editorial note*]
+
+#### `execution::bulk_execute`
 
-### `BulkOneWayExecutor` requirements
+The name `execution::bulk_execute` denotes a customization point object. If `is_convertible_v<S, executor_shape_t<remove_cvref_t<E>>>` is true, then the expression `execution::bulk_execute(E, F, S)` for some subexpressions `E`, `F`, and `S` is expression-equivalent to:
```
LeeHowes: We should probably use the letter S here instead of E and refer to Senders generally, as that's the interface this depends on rather than Executor.

wording.md
```diff
@@ -121,55 +138,211 @@ namespace execution {
...
+- Otherwise, `value(C, V)`, if that expression is valid, with overload resolution performed in a context that includes the declaration
 
-An executor type's destructor shall not block pending completion of the submitted function objects. [*Note:* The ability to wait for completion of submitted function objects may be provided by the associated execution context. *--end note*]
+        template<class C, class V>
+          void value(C, V) = delete;
```
ericniebler:
```suggestion
          void value();
```

wording.md
```diff
@@ -121,55 +138,211 @@ namespace execution {
...
+### Concept `callback`
+
+XXX TODO The `callback` concept...
+
+```
+template<class T, class... An>
+concept callback =
+  invocable<T, An...> &&
+  callback_signal<T>;
+```
```
ericniebler: In my C++Con talk, I called `callback_signal` "`callback`", and I called `callback` "`callback_of`". I liked the ergonomics of that. That's a personal preference, and not a hill to die on.

wording.md
```diff
@@ -121,55 +138,211 @@ namespace execution {
...
+
+### Concept `sender_to`
+
+XXX TODO The `sender_to` concept...
+
+```
+template<class S, class C>
+concept sender_to =
```
ericniebler:
```suggestion
concept sender_to =
  callback_signal<C> &&
```

wording.md
```diff
@@ -121,55 +138,211 @@ namespace execution {
...
+
+[*Editorial note:* This specification is adapted from `ranges::iter_swap`. *--end editorial note*]
+
+### Concept `callback_signal`
+
+XXX TODO The `callback_signal` concept...
+
+```
+template<class T, class E = exception_ptr>
+concept callback_signal =
```
ericniebler:
```suggestion
concept callback_signal =
  (nothrow_?)move_constructible<remove_cvref_t<T>> &&
```

wording.md
```diff
@@ -121,55 +138,211 @@ namespace execution {
...
+  (
+    requires(E&& e, F&& f) {
+      execution::execute((E&&)e,(F&&)f);
+    }
+
+    ||
+
+    requires(E&&e, F&& f) {
+      execution::submit((E&&)e,(F&&)f);
+    }
```
ericniebler: Strongly lobby to strike this disjunction here.

wording.md
```diff
@@ -121,55 +138,211 @@ namespace execution {
...
+  );
+  };
+```
+
+None of an executor's copy constructor, destructor, equality comparison, or `swap` operation shall exit via an exception.
+
+None of these operations, nor an executor type's `execute` or `submit` functions, or associated query functions shall introduce data races as a result of concurrent invocations of those functions from different threads.
+
+For any two (possibly const) values `x1` and `x2` of some executor type `X`, `x1 == x2` shall return `true` only if `x1.query(p) == x2.query(p)` for every property `p` where both `x1.query(p)` and `x2.query(p)` are well-formed and result in a non-void type that is `EqualityComparable` (C++Std [equalitycomparable]). [*Note:* The above requirements imply that `x1 == x2` returns `true` if `x1` and `x2` can be interchanged with identical effects. An executor may conceptually contain additional properties which are not exposed by a named property type that can be observed via `execution::query`; in this case, it is up to the concrete executor implementation to decide if these properties affect equality. Returning `false` does not necessarily imply that the effects are not identical. *--end note*]
```
ericniebler: s/`EqualityComparable`/`equality_comparable`/. My earlier comment seems to be lost. I think this can probably be expressed more precisely with, "For any (possibly) const value `x` of executor type `X` and `p` of property type `P`, the expression `x.query(p)`, if it is well-formed, is equality-preserving ([concepts.equality]).

wording.md
jaredhoberock: Done

wording.md
```diff
@@ -121,55 +138,211 @@ namespace execution {
...
+| `e.submit(f)` | `void` | Evaluates `DECAY_COPY(std::forward<F>(f))` on the calling thread to create `cf` that will be invoked at most once by an execution agent. <br/> May block pending completion of this invocation. <br/> Synchronizes with [intro.multithread] the invocation of `f`. <br/>Shall not propagate any exception thrown by the function object or any other function submitted to the executor. [*Note:* The treatment of exceptions thrown by one-way submitted functions is described by the `execution::oneway_exception_handler` property. The forward progress guarantee of the associated execution agent(s) is implementation defined. *--end note.*] |
```
ericniebler: Once we remove the disjunction from the concept `executor`, and the ability to pass an `invocable` to `submit`, this gets simpler and can be relocated to be the semantics of the `submit` customization point.

wording.md
```diff
@@ -121,55 +138,211 @@ namespace execution {
...
 
-| Expression | Return Type | Operational semantics |
-|------------|-------------|---------------------- |
-| `x.execute(f)` | `void` | Evaluates `DECAY_COPY(std::forward<F>(f))` on the calling thread to create `cf` that will be invoked at most once by an execution agent. <br/> May block pending completion of this invocation. <br/> Synchronizes with [intro.multithread] the invocation of `f`. <br/>Shall not propagate any exception thrown by the function object or any other function submitted to the executor. [*Note:* The treatment of exceptions thrown by one-way submitted functions is described by the `execution::oneway_exception_handler` property. The forward progress guarantee of the associated execution agent(s) is implementation defined. *--end note.*] |
+- `C.done()`, if that expression is valid. If the function selected does not signal the callback `C`'s done channel, the program is ill-formed with no diagnostic required.
+
+- Otherwise, `done(C)`, if that expression is valid, with overload resolution performed in a context that includes the declaration
+
+        template<class C>
+          void done(C) = delete;
```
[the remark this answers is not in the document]
jaredhoberock: Should this be

    template<class C>
      void done();

or simply

      void done();

?

And others similarly?

wording.md
jaredhoberock: Done

wording.md
LeeHowes: Without the invocable part though is callback something with value, error, done, and callback_of one that exposes the type list that value can send?

wording.md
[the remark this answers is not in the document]
ericniebler: Just `void done();`. Poison-pill overloads exist to (a) seed the overload set with something non-viable, and (b) create an intentional ambiguity with any pre-existing unconstrained function templates of the same name that are likely to be found by ADL.

The template `iter_swap` poison-pill in `std::ranges` owes its existence to the wildly under-constrained `std::iter_swap`, which will happy gobble up everything if ADL considers it. None of these customization points have the need to create an ambiguity with such an overload because none exist (thankfully). So we only need to seed the overload set with something non-viable. `void done();` suffices.

wording.md
```diff
@@ -121,55 +138,211 @@ namespace execution {
...
+#### `execution::submit`
+
+The name `execution::submit` denotes a customization point object. The expression `execution::submit(S, C)` for some subexpressions `S` and `C` is expression-equivalent to:
+
+- `S.submit(C)`, if that expression is valid. If the function selected does not submit the callback object `C` via the sender `S`, the program is ill-formed with no diagnostic required.
+
+- Otherwise, `submit(S, C)`, if that expression is valid, with overload resolution performed in a context that includes the declaration
+
+        template<class S, class C>
+          void submit(S, C) = delete;
```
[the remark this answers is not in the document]
jaredhoberock: Done

wording.md
```diff
@@ -121,55 +138,211 @@ namespace execution {
...
+#### `execution::execute`
+
+The name `execution::execute` denotes a customization point object. The expression `execution::execute(E, F)` for some subexpressions `E` and `F` is expression-equivalent to:
+
+- `E.execute(F)`, if that expression is valid. If the function selected does not execute the function object `F` on the executor `E`, the program is ill-formed with no diagnostic required.
+
+- Otherwise, `execute(E, F)`, if that expression is valid, with overload resolution performed in a context that includes the declaration
+
+        template<class E, class F>
+          void execute(E, F) = delete;
```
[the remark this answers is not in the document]
jaredhoberock: Done

wording.md
```diff
@@ -121,55 +138,211 @@ namespace execution {
...
+#### `execution::error`
+
+The name `execution::error` denotes a customization point object. The expression `execution::error(C, E)` for some subexpressions `C` and `E` are expression-equivalent to:
+
+- `C.error(E)`, if that expression is valid. If the function selected does not send the error `E` to the callback `C`'s error channel, the program is ill-formed with no diagnostic required.
+
+- Otherwise, `error(C, E)`, if that expression is valid, with overload resolution performed in a context that includes the declaration
+
+        template<class C, class E>
+          void error(C, E) = delete;
```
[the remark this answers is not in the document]
jaredhoberock: Done

wording.md
```diff
@@ -19,6 +19,36 @@ For the intent of this library and extensions to this library, the *lifetime of
...
 namespace execution {
 
+  // Customization points:
+
+  inline namespace unspecified{
+    inline constexpr unspecified value = unspecified;
+
+    inline constexpr unspecified done = unspecified;
+
+    inline constexpr unspecified error = unspecified;
```
lewissbaker: In P1660 these operations were member functions.
In other examples where we have made these CPOs we have named these with the `set_` prefix.
Should these CPOs be named: `set_value`, `set_done` and `set_error`?

jaredhoberock:
Merging this WIP

PR #441, opened 2019-10-05
Rename callback back to receiver, operator() back to set_value…

ericniebler:
…introduce sender concept and sink_receiver

### DIFF
--- a/wording.md
+++ b/wording.md
@@ -22,11 +22,11 @@ namespace execution {
   // Customization points:
 
   inline namespace unspecified{
-    inline constexpr unspecified value = unspecified;
+    inline constexpr unspecified set_value = unspecified;
 
-    inline constexpr unspecified done = unspecified;
+    inline constexpr unspecified set_done = unspecified;
 
-    inline constexpr unspecified error = unspecified;
+    inline constexpr unspecified set_error = unspecified;
 
     inline constexpr unspecified execute = unspecified;
 
@@ -38,16 +38,22 @@ namespace execution {
   // Concepts:
 
   template<class T, class E = exception_ptr>
-    concept callback_signal = see-below;
+    concept receiver = see-below;
 
   template<class T, class... an>
-    concept callback = see-below;
+    concept receiver_of = see-below;
+
+  template<class S>
+    concept sender = see-below;
 
   template<class S, class C>
     concept sender_to = see-below;
 
   template<class E, class F = void(*)()>
     concept executor = see-below;
+
+  // A no-op receiver type
+  class sink_receiver;
 
   // Indication of executor property applicability
   template<class T> struct is_executor;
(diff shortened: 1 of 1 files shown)


wording.md
```diff
@@ -254,42 +260,60 @@ The name `execution::bulk_execute` denotes a customization point object. If `is_
...
 
 ```
 template<class T, class... An>
-concept callback =
-  invocable<T, An...> &&
-  callback_signal<T>;
+concept receiver_of =
+  receiver<T> &&
+  requires(T&& t, An&&... an) {
+    execution::set_value((An&&) an...);
```
kirkshoop: Add t to set_value exp

wording.md
```diff
@@ -140,57 +146,57 @@ A type `A` meets the `ProtoAllocator` requirements if `A` is `CopyConstructible`
 
 ### Customization points
 
-#### `execution::value`
+#### `execution::set_value`
 
-The name `execution::value` denotes a customization point object. The expression `execution::value(C, V)` for some subexpressions `C` and `V` is expression-equivalent to:
+The name `execution::set_value` denotes a customization point object. The expression `execution::set_value(R, Vs...)` for some subexpressions `R` and `Vs...` is expression-equivalent to:
```
lewissbaker: Do we need to define `Vs...` to be some comma-separated list of zero or more arguments?
Or is it clear enough that this is a variadic function call?

ericniebler: I'm not spending time getting super picky with the wording. LWG can fix it how they want. They will anyway.

wording.md
```diff
@@ -216,21 +222,21 @@ The name `execution::execute` denotes a customization point object. The expressi
...
-- Otherwise, `submit(S, C)`, if that expression is valid, with overload resolution performed in a context that includes the declaration
+- Otherwise, `submit(S, R)`, if that expression is valid, with overload resolution performed in a context that includes the declaration
 
         void submit();
 
-    and that does not include a declaration of `execution::submit`. If the function selected by overload resolution does not submit the callback object `C` via the sender `S`, the program is ill-formed with no diagnostic required.
+    and that does not include a declaration of `execution::submit`. If the function selected by overload resolution does not submit the receiver object `R` via the sender `S`, the program is ill-formed with no diagnostic required.
 
-- Otherwise, if the type `C` models `Callback`, then let `F = [C]{ C.value(); }`. If `S` and `F` model `executor`, then `execution::execute(S, F)`.
+- Otherwise, if the type `R` models `receiver`, then let `F = [R]{ R.set_value(); }`. If `S` and `F` model `executor`, then `execution::execute(S, F)`.
```
lewissbaker: This needs some wording to describe an adapter that calls `set_done()` in its destructor if `set_value()` is not called and the object has not been moved-from, etc.

lewissbaker: Should `submit(s, r)` always require that `r` models `receiver`?
Or are we allowing any old thing to be passed in here as long as there is a submit customisation defined for that combination?

ericniebler: Agree we should require that.

wording.md
```diff
@@ -323,14 +346,29 @@ In addition to the above requirements, types `E` and `F` model `executor` only i
...
-- `f` denotes a function or callback object of type `F&&` invocable as `cf()` and where `decay_t<F>` satisfies the `MoveConstructible` requirements.
+- `cf` denotes the function or receiver object `DECAY_COPY(std::forward<F>(f))` 
+- `f` denotes a function or receiver object of type `F&&` invocable as `cf()` and where `decay_t<F>` models `move_constructible`.
 
 | Expression | Return Type | Operational semantics |
 |------------|-------------|---------------------- |
 | `e.execute(f)` | `void` | Evaluates `DECAY_COPY(std::forward<F>(f))` on the calling thread to create `cf` that will be invoked at most once by an execution agent. <br/> May block pending completion of this invocation. <br/> Synchronizes with [intro.multithread] the invocation of `f`. <br/>Shall not propagate any exception thrown by the function object or any other function submitted to the executor. [*Note:* The treatment of exceptions thrown by one-way submitted functions is described by the `execution::oneway_exception_handler` property. The forward progress guarantee of the associated execution agent(s) is implementation defined. *--end note.*] |
```
lewissbaker: Interesting that the wording here _requires_ that the function object be copied using DECAY_COPY and that that copy is invoked. This forces a copy to be taken, even if the executor is the inline_executor. Should the wording be relaxed to allow the implementation to _optionally_ copy the object using DECAY_COPY?

jaredhoberock: Not sure what the correct course to take is, but IIRC, we copy/pasted similar wording from `std::thread`'s ctor and `std::async`.

jaredhoberock:
Thanks Eric & Lewis, is this ready to merge?

jaredhoberock:
I'm going to merge these pull requests now to reduce the possibility of introducing conflicts with my changes.

PR #444, opened 2019-10-05
WIP: Miscellaneous edits

jaredhoberock:


wording.md
```diff
@@ -8,7 +8,7 @@ This Clause describes components supporting execution of function objects [funct
...
+An execution agent invokes a function object within an *execution context* such as the calling thread or thread-pool.  An *executor* submits a function object to an execution context to be invoked by an execution agent within that execution context. [*Note:* Invocation of the function object may be inlined such as when the execution context is the calling thread, or may be scheduled such as when the execution context is a thread-pool with task scheduler. *--end note*] An executor may submit a function object with *execution properties* that specify how the submission and invocation of the function object interacts with the submitting thread and execution context, including forward progress guarantees [intro.progress].
```
jaredhoberock: This change is just whitespace

wording.md
```diff
@@ -207,27 +220,45 @@ The name `execution::set_error` denotes a customization point object. The expres
...
+          using invocable_type = std::remove_cvref_t<F>;
+          invocable_type f_;
+        public:
+          explicit as-receiver(invocable_type&& f) : f_(move_if_noexcept(f)) {}
+          explicit as-receiver(const invocable_type& f) : f_(f) {}
+          as-receiver(as-receiver&& other) = default;
+          void set_value() {
+            invoke(f_);
+          }
+          void set_error(std::exception_ptr) noexcept {}
```
kirkshoop: I expect set_error to call `terminate()`

errors should not be ignored.

wording.md
```diff
@@ -207,27 +220,45 @@ The name `execution::set_error` denotes a customization point object. The expres
...
+        struct as-receiver {
+        private:
+          using invocable_type = std::remove_cvref_t<F>;
+          invocable_type f_;
+        public:
+          explicit as-receiver(invocable_type&& f) : f_(move_if_noexcept(f)) {}
+          explicit as-receiver(const invocable_type& f) : f_(f) {}
+          as-receiver(as-receiver&& other) = default;
+          void set_value() {
+            invoke(f_);
```
AerialMantis: `invoke` -> `std::invoke`.

wording.md
```diff
@@ -304,23 +331,26 @@ The name `execution::schedule` denotes a customization point object. The express
...
 
-- Otherwise, `execution::bulk_execute(E, F, S)` is ill-formed.
+- Otherwise, if the types `F` and `executor_index_t<remove_cvref_t<S>>` model `invocable` and if `execution::query(S, execution::bulk_guarantee)` equals `execution::bulk_guarantee.unsequenced`, then
+    - Evaluates `DECAY_COPY(std::forward<decltype(F)>(F))` on the calling thread to create a function object `cf`. [*Note:* Additional copies of `cf` may subsequently be created. *--end note.*]
+    - For each value of `i` in `N`, `cf(i)` (or copy of `cf`)) will be invoked at most once by an execution agent that is unique for each value of `i`.
+    - May block pending completion of one or more invocations of `cf`.
+    - Synchronizes with (C++Std [intro.multithread]) the invocations of `cf`.
+    - Shall not propagate any exception thrown by `cf` or any other function submitted to the executor. [*Note:* The treatment of exceptions thrown by `bulk_execute`-submitted functions are implementation-defined. *--end note.*]
```
kirkshoop: 'Shall not propagate' seems a bit too constrained. it is okay for the implementation to call receiver.set_error once with an error or an aggregate of all errors..

wording.md
```diff
@@ -415,22 +445,30 @@ In the Table below,
...
 ```
-template<class E, class F = void(*)()>
-concept executor =
-  invocable<F> &&
-  copy_constructible<remove_cvref_t<E>> &&
-  equality_comparable<remove_cvref_t<E>> &&
+template<class E, class F>
+concept executor-impl =
+  invocable<F>&&
+  equality_comparable<E> &&
```
kirkshoop: I expected this to be nothrow move and nothrow copy based on the text "None of an executor's copy constructor, destructor, equality comparison, or `swap` operation shall exit via an exception."

wording.md
```diff
@@ -896,7 +934,7 @@ template<class Executor>
   friend constexpr S query(const Executor& ex, const Property& p) noexcept(noexcept(execution::query(ex, std::declval<const S::Nk>())));
 ```
 
-*Returns:* `std::query(ex, S::N`*k*`())`.
+*Returns:* `execution::query(ex, S::N`*k*`())`.
```
kirkshoop: I think that properties is now in std::

wording.md
```diff
@@ -207,27 +220,45 @@ The name `execution::set_error` denotes a customization point object. The expres
...
-- Otherwise, `execute(E, F)`, if that expression is valid, with overload resolution performed in a context that includes the declaration
+- Otherwise, `execute(e, f)`, if that expression is valid, with overload resolution performed in a context that includes the declaration
 
         void execute();
 
-    and that does not include a declaration of `execution::execute`. If the function selected by overload resolution does not execute the function object `F` on the executor `E`, the program is ill-formed with no diagnostic required.
+    and that does not include a declaration of `execution::execute`. If the function selected by overload resolution does not execute the function object `f` on the executor `e`, the program is ill-formed with no diagnostic required.
 
-- Otherwise, if the type `F` models `invocable`, then `execution::submit(E, execution::make_callback_from(F))`.
+- Otherwise, `execution::submit(e, `_`as-receiver`_`<F>(forward<F>(f)))` if `E` models `sender`, where _`as-receiver`_ is some implementation-defined class template equivalent to:
```
AerialMantis: Should we also have a requirement here that `E` model `sender_of<as-receiver<F>>`, otherwise I imagine this will be ill-formed?

wording.md
```diff
@@ -54,9 +58,12 @@ namespace execution {
   template<class S>
     concept scheduler = see-below;
 
-  template<class E, class F = void(*)()>
+  template<class E>
```
AerialMantis: Now that the `executor` concept is in terms of `E` only, we will also need to update the concept definition for `executor` and add a definition for `executor_of`.

I think the current contract requirements in the concept definition for `executor` may need to move to `executor_of` as they rely on knowledge of the function type as well as the executor type.

These changes may conflict slightly with some minor edits I proposed in #446

wording.md
```diff
@@ -304,23 +331,26 @@ The name `execution::schedule` denotes a customization point object. The express
 
 #### `execution::bulk_execute`
 
-The name `execution::bulk_execute` denotes a customization point object. If `is_convertible_v<S, executor_shape_t<remove_cvref_t<E>>>` is true, then the expression `execution::bulk_execute(E, F, S)` for some subexpressions `E`, `F`, and `S` is expression-equivalent to:
+The name `execution::bulk_execute` denotes a customization point object. If `is_convertible_v<N, size_t>` is true, then the expression `execution::bulk_execute(S, F, N)` for some subexpressions `S`, `F`, and `N` is expression-equivalent to:
```
AerialMantis: Should we use `E` to denote the executor in these expressions?

wording.md
```diff
@@ -304,23 +331,26 @@ The name `execution::schedule` denotes a customization point object. The express
...
 
-    and that does not include a declaration of `execution::bulk_execute`. If the function selected by overload resolution does not execute `S` invocations of the function object `F` on the executor `E` in bulk, and the result of that function does not model `sender<void>`, the program is ill-formed with no diagnostic required.
-
-- Otherwise, if the types `F` and `executor_index_t<remove_cvref_t<E>>` model `invocable`, then `execution::tbd_concrete_implementations::bulk_execute(E, F, S)`.
+    and that does not include a declaration of `execution::bulk_execute`. If the function selected by overload resolution does not execute `N` invocations of the function object `F` on the executor `S` in bulk with forward progress guarantee `execution::query(E, execution::bulk_guarantee)`, and the result of that function does not model `sender<void>`, the program is ill-formed with no diagnostic required.
 
-- Otherwise, `execution::bulk_execute(E, F, S)` is ill-formed.
+- Otherwise, if the types `F` and `executor_index_t<remove_cvref_t<S>>` model `invocable` and if `execution::query(S, execution::bulk_guarantee)` equals `execution::bulk_guarantee.unsequenced`, then
```
AerialMantis: I believe this wording will restrict the adaptation of `execution::bulk_execute` to when the executor has `execution::bulk_guarantee.unsequenced`. Should we provide alternative adaptations for the other bulk guarantee properties, or add a note to mention that we will in the future?

wording.md
jaredhoberock: Done

wording.md
jaredhoberock: Wording doesn't usually qualify calls to functions in `std`

wording.md
jaredhoberock: Lee asked for `S`

wording.md
jaredhoberock: I think an executor should only provide `bulk_guarantee` when it also provides a specialization of `bulk_execute`. I don't think it makes sense for a fallback implementation of `bulk_execute` to attempt to provide a stronger forward progress guarantee than unsequenced.

wording.md
jaredhoberock: I've removed that bullet

wording.md
jaredhoberock: I've added `is_nothrow_copy_constructible` & `is_nothrow_destructible` to `executor-impl`

wording.md
jaredhoberock: Fixed

wording.md
jaredhoberock: Fixed

wording.md
AerialMantis: That's fine, this was just for consistency because I saw `std` used in other parts of the definition.

wording.md
AerialMantis: This doesn't seem consistent with the rest of the paper, we generally use `S` to denote a sender type and `E` to denote an executor type.

wording.md
AerialMantis: Okay yeah, that makes sense.

wording_front_matter.md
```diff
@@ -57,26 +57,22 @@ Abstract:           This paper proposes a programming model for executors, which
...
-As directed by SG1 at the 2019-XX Cologne meeting, we have implemented the following changes suggested by P1658 and P1660:
+As directed by SG1 at the 2019-07 Cologne meeting, we have implemented the following changes suggested by P1658 and P1660:
 
 * Eliminated interface-changing properties `oneway_t` and `bulk_oneway_t`.
-* Introduced `executor` concept.
+* Introduced `executor` and `executor_of` concepts.
 * Eliminated `OneWayExecutor` and `BulkOneWayExecutor` requirements.
-* Eliminated `is_oneway_executor` and `is_bulk_oneway_executor` type traits.
+* Eliminated `is_executor`, `is_oneway_executor`, and `is_bulk_oneway_executor` type traits.
 * Introduced `callback_signal`, `callback`, and `sender_to` concepts.
```
kirkshoop: switch callback to receiver?

wording_front_matter.md
```diff
@@ -57,26 +57,22 @@ Abstract:           This paper proposes a programming model for executors, which
 
 ### Revision 11
 
-As directed by SG1 at the 2019-XX Cologne meeting, we have implemented the following changes suggested by P1658 and P1660:
+As directed by SG1 at the 2019-07 Cologne meeting, we have implemented the following changes suggested by P1658 and P1660:
 
 * Eliminated interface-changing properties `oneway_t` and `bulk_oneway_t`.
-* Introduced `executor` concept.
```
kirkshoop: add scheduler, sender concepts

wording_front_matter.md
```diff
@@ -57,26 +57,22 @@ Abstract:           This paper proposes a programming model for executors, which
...
+* Introduced `executor` and `executor_of` concepts.
 * Eliminated `OneWayExecutor` and `BulkOneWayExecutor` requirements.
-* Eliminated `is_oneway_executor` and `is_bulk_oneway_executor` type traits.
+* Eliminated `is_executor`, `is_oneway_executor`, and `is_bulk_oneway_executor` type traits.
 * Introduced `callback_signal`, `callback`, and `sender_to` concepts.
 * Introduced `value`, `error`, `done`, `execute`, `submit`, and `bulk_execute` customization point objects.
```
kirkshoop: set_value, set_error, set_done and schedule

wording.md
```diff
@@ -414,28 +432,83 @@ In the Table below,
...
+
+```
+template<class E>
+concept executor = executor-impl<E, execution::invocable_archetype>;
+```
+
+Neither of an executor's equality comparison or `swap` operation shall exit via an exception.
 
-None of these operations, nor an executor type's `execute` or `submit` functions, or associated query functions shall introduce data races as a result of concurrent invocations of those functions from different threads.
+None of an executor type's copy constructor, destructor, equality comparison, `swap` function, `execute` function, `submit` functions, or associated `query` functions shall introduce data races as a result of concurrent invocations of those functions from different threads.
```
kirkshoop: remove submit reference

wording.md
jaredhoberock: Yes, but Lee pointed out that this is a sender algorithm

AerialMantis: Does that mean that this algorithm should expect a sender type? Because, the requirements seem to expect an executor type.

jaredhoberock:
Merging this now to avoid additional conflicts

PR #445, opened 2019-10-05
Polymorphic executor fixes

chriskohlhoff:


jaredhoberock:
Thanks for these fixes, this stuff slipped my mind on the call.

PR #447, opened 2019-10-06
add typed_sender concept

ericniebler:
This concept is important to the hpc folks because it enables them to discover whether a sender is "no-fail" (does not ever call `set_error` or `set_done` on a receiver).

It is also important to Facebook because it makes it possible to automatically adapt a sender to be awaitable in a coroutine.

### DIFF
--- a/wording.md
+++ b/wording.md
@@ -48,6 +48,9 @@ namespace execution {
   template<class S>
     concept sender = see-below;
 
+  template<class S>
+    concept typed_sender = see-below;
+
   template<class S, class R>
     concept sender_to = see-below;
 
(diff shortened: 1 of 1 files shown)


jaredhoberock:
Should we require that `bulk_execute` return a `typed_sender` of `void`? Its wording currently says something like `sender<void>`, which I assume is nonsense.

ericniebler:
It would be nice to require `typed_sender` where the value, error, and done channels are knowable _a priori_. Is this the case for `bulk_execute`?

It may turn out we need some concept like `typed_sender_of<value-types...>`, but how to also specify the error and done channel there is an interesting question.

AerialMantis:
I think this would be useful to have, perhaps we could have separate concepts for this, something like `sender_of_values<value-types...>` and `sender_of_errors<error-types...>`?

Would we also need this for the `done` channel or is it expected that `done` doesn't take a parameter?

PR #455, opened 2020-01-24
typos and missing stuff

menuet:
Hi,
While trying to implement the concepts in the paper in order to better understand them, I found some things in the wording that seem to be typos or missing pieces.
I propose this pull request for you to review and fix in the way you want.
Because I only understand things partially, these may be false issues, in this case, sorry for loosing your time :-)

### DIFF
--- a/wording.md
+++ b/wording.md
@@ -241,7 +241,7 @@ For some subexpressions `e` and `f`, let `E` be a type such that `decltype((e))`
           void set_value() {
             invoke(f_);
           }
-          void set_error(std::exception_ptr) {
+          void set_error(std::exception_ptr) noexcept {
             terminate();
           }
           void set_done() noexcept {}
@@ -444,7 +444,7 @@ template<template<class...> class Variant>
 struct has-error-types; // exposition only
 
 template<class S>
-comcept has-sender-types = // exposition only
+concept has-sender-types = // exposition only
   requires {
     typename has-value-types<S::template value_types>;
     typename has-error-types<S::template error_types>;
@@ -467,7 +467,7 @@ concept scheduler =
   copy_constructible<remove_cvref_t<S>> &&
   equality_comparable<remove_cvref_t<S>> &&
   requires(E&& e) {
-    execution::schedule((S&&)s);
+    execution::schedule((E&&)e);
   }; // && sender<invoke_result_t<execution::schedule, S>>
 ```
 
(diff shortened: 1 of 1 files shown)


jaredhoberock:
Thanks for the review, and for taking a stab at producing an implementation! We'll fix these bugs in the next revision.

menuet:
I am glad if my findings are real and if they can help you.
By the way, there is another thing I am struggling with:
When I read the spec, there is a kind of infinite recursion in the defintion of some concepts.
For instance, the concept _executor_ depends on the niebloid _execution::execute_, but the implementation of _execution::execute_ needs to select some overload based on the concept _executor_.
Maybe you could give me a hint at how I need to implement this? Do I read it incorrectly?

jaredhoberock:
> For instance, the concept `executor` depends on the niebloid `execution::execute`, but the implementation of `execution::execute` needs to select some overload based on the concept `executor`.

I think this is not quite correct. It's true that concept `executor` checks that `execution::execute` is well-formed. That is, concept `executor` is defined in terms of `execution::execute`.

However, `execution::execute` only checks for either a member `.execute` or a free function `execute`. So `execution::execute` should be terminal. At least, that's the intended definition.

Here is how I implemented `execution::execute` (in C++11):

```
struct execute_customization_point
{
  template<class E, class F,
           P0443_REQUIRES(has_execute_member_function<E&&,F&&>::value)
          >
  constexpr auto operator()(E&& e, F&& f) const ->
    decltype(std::forward<E>(e).execute(std::forward<F>(f)))
  {
    return std::forward<E>(e).execute(std::forward<F>(f));
  }

  template<class E, class F,
           P0443_REQUIRES(!has_execute_member_function<E&&,F&&>::value and
                          has_execute_free_function<E&&,F&&>::value)
          >
  constexpr auto operator()(E&& e, F&& f) const ->
    decltype(execute(std::forward<E>(e), std::forward<F>(f)))
  {
    return execute(std::forward<E>(e), std::forward<F>(f));
  }

  template<class E, class F,
           P0443_REQUIRES(!has_execute_member_function<E&&,F&&>::value and
                          !has_execute_free_function<E&&,F&&>::value and
                          has_submit_member_function<E&&,invocable_as_receiver<F&&>>::value)
          >
  constexpr auto operator()(E&& e, F&& f) const ->
    decltype(std::forward<E>(e).submit(detail::as_receiver(std::forward<F>(f))))
  {
    // note that this overload does not use execution::submit to avoid circular dependency
    return std::forward<E>(e).submit(detail::as_receiver(std::forward<F>(f)));
  }

  template<class E, class F,
           P0443_REQUIRES(!has_execute_member_function<E&&,F&&>::value and
                          !has_execute_free_function<E&&,F&&>::value and
                          !has_submit_member_function<E&&,invocable_as_receiver<F&&>>::value and
                          has_submit_free_function<E&&,invocable_as_receiver<F&&>>::value)
          >
  constexpr auto operator()(E&& e, F&& f) const ->
    decltype(submit(std::forward<E>(e), detail::as_receiver(std::forward<F>(f))))
  {
    // note that this overload does not use execution::submit to avoid circular dependency
    return submit(std::forward<E>(e), detail::as_receiver(std::forward<F>(f)));
  }
};
```

menuet:
OK. Thanks a lot for your example. This is really helpfull.
The sentence in the wording that made me thinking about recursion is:
 _The expression execution::execute(e, f) is ill-formed if F does not model invocable, or if E does not model either executor or sender_
So, at first, I thought that, inside execution::execute, I had to check explicitely that E models executor (or sender).
In fact, you only check against smaller constraints in order to select the proper overload.
Thanks again.
Cor3ntin also pointed me to an implementation from Eric Niebler, so now I have 2 more things to read and analyze :-)

jaredhoberock:
Eric's PR fixed most of these issues out-of-band.

PR #458, opened 2020-02-24
merge P2006R1 into P0443, renumber as R13

ericniebler:


wording.md
```diff
@@ -352,124 +439,113 @@ The name `execution::bulk_execute` denotes a customization point object. If `is_
...
-concept scheduler =
-  copy_constructible<remove_cvref_t<S>> &&
-  equality_comparable<remove_cvref_t<S>> &&
-  requires(E&& e) {
-    execution::schedule((S&&)s);
-  }; // && sender<invoke_result_t<execution::schedule, S>>
-```
+        template<class S>
+          concept scheduler =
+            copy_constructible<remove_cvref_t<S>> &&
```
lewissbaker: With the `executor` concept we require `is_nothrow_copy_constructible_v<E>`.
Should we also be requiring that of the scheduler concept?

Perhaps also `is_nothrow_move_constructible_v<E>`?

wording.md
```diff
@@ -352,124 +439,113 @@ The name `execution::bulk_execute` denotes a customization point object. If `is_
...
-  has-sender-types<sender_traits<S>>;
-```
+        template<class S>
+          concept typed_sender =
+            sender<S> &&
+            has-sender-types<sender_traits<remove_cvref_t<S>>>;
 
 ### Concept `scheduler`
 
 XXX TODO The `scheduler` concept...
```
lewissbaker: What is still TODO here?

wording.md
```diff
@@ -540,27 +616,28 @@ XXX TODO The `executor` and `executor_of` concepts...
...
 
-template<class E, class F>
-concept executor_of = executor-of-impl<E, F>;
-```
+        template<class E, class F>
+          concept executor_of =
+            executor<E> &&
+            executor-of-impl<E, F>;
 
 Neither of an executor's equality comparison or `swap` operation shall exit via an exception.
```
lewissbaker: This references an executor's `swap` operation but the `executor` concept does not require `swappable`.
Should it?

wording.md
```diff
@@ -609,30 +671,71 @@ The class template `sender_traits` can be used to query information about a `sen
...
+
+- Otherwise, let _`void-receiver`_ be an implementation-defined class type equivalent to
+
+        struct void-receiver { // exposition only
+          void set_value() noexcept;
+          void set_error(exception_ptr) noexcept;
+          void set_done() noexcept;
+        };
+
+    If _`executor-of-impl`_`<S, `_`as-invocable`_`<`_`void-receiver`_`, S>>` is `true`, then _`sender-traits-base`_ is equivalent to
```
lewissbaker: Could this requirement be simplified to `executor<S>` and avoid the need for `void-receiver`?

wording.md
```diff
@@ -227,108 +247,175 @@ For some subexpressions `e` and `f`, let `E` be a type such that `decltype((e))`
...
-- Otherwise, `schedule(S)`, if that expression is valid and its type `N` models `sender`
+- Otherwise, `schedule(s)`, if that expression is valid and its type models `sender`
   with overload resolution performed in a context that includes the declaration
 
         void schedule();
 
     and that does not include a declaration of `execution::schedule`. 
 
-- Otherwise, _`decay-copy`_`(S)` if the type `S` models `sender`.
+- Otherwise, _`as-sender`_`<remove_cvref_t<S>>{s}` if `S` satisfies `executor`, where _`as-sender`_ is an
```
lewissbaker: Do we need the `as-sender` adapter here?
An `executor` already has a default implementation of `execution::connect()` and has a specialisation of `sender_traits` that reports the correct `value_types`, `error_types`, etc.

I'm not sure what wrapping the executor up in a sender-wrapper is doing for us.

wording.md
```diff
@@ -540,27 +616,28 @@ XXX TODO The `executor` and `executor_of` concepts...
...
-    execution::execute(e, (F&&)f);
-  };
-```
+        template<class E, class F>
+          concept executor-of-impl =
+            invocable<remove_cvref_t<F>&> &&
+            constructible_from<remove_cvref_t<F>, F> &&
+            move_constructible<remove_cvref_t<F>> &&
+            copy_constructible<E> &&
+            is_nothrow_copy_constructible_v<E> &&
```
lewissbaker: We are requiring `is_nothrow_copy_constructible_v<E>` but not requiring `is_nothrow_move_constructible_v<E>`.
Should we also add a requirement for `is_nothrow_move_constructible_v<E>`?

P0443-syntax.xml
```diff
@@ -21,6 +21,7 @@
       <item>blocking</item>
       <item>blocking_t</item>
       <item>bulk_execute</item>
+      <item>connect</item>
```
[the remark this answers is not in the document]
lewissbaker: Also `sender_traits` and `sender_base`?

ericniebler:
@jaredhoberock is it within scope to fix bugs here and now, or do we need to merge P2006 as SG1 saw it and go through a process to make additional changes?

jaredhoberock:
@ericniebler I think it depends on the issue. We can fix bugs that are obvious oversights but for edge cases we should wait for SG1 & LEWG's wording reviews and raise those issues there.

In SG1's case, they will comb over the wording in full committee.

wording.md
ericniebler: No, because we need the _`as-invocable`_ here to keep this from recursing.

wording.md
```diff
@@ -227,108 +247,175 @@ For some subexpressions `e` and `f`, let `E` be a type such that `decltype((e))`
...
-          explicit as-receiver(invocable_type&& f) : f_(move_if_noexcept(f)) {}
-          explicit as-receiver(const invocable_type& f) : f_(f) {}
-          as-receiver(as-receiver&& other) = default;
-          void set_value() {
+          F f_;
+          void set_value() noexcept(is_nothrow_invocable_v<F&>) {
             invoke(f_);
           }
-          void set_error(std::exception_ptr) {
+          [[noreturn]] void set_error(std::exception_ptr) noexcept {
```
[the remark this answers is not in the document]
lewissbaker: `as-receiver` is used in places where a void-sender is used as an executor (ie. the sender is passed to `execution::execute(s, f)`).
Thus it's possible that the sender, `s`, might send any error type, not just `exception_ptr`, and so the `as-receiver` should allow `set_error()` to be invoked with any error type and map them all to `std::terminate()`.

wording.md
```diff
@@ -227,108 +247,175 @@ For some subexpressions `e` and `f`, let `E` be a type such that `decltype((e))`
...
 
-    and that does not include a declaration of `execution::submit`. If the function selected by overload resolution does not submit the receiver object `r` via the sender `s`, the program is ill-formed with no diagnostic required.
+        struct as-operation {
+          U e_;
+          T r_;
+          void start() noexcept try {
+            execution::execute(std::move(e_), as-invocable <T, S>{r_});
+          } catch(...) {
+            execution::set_error(std::move(r_), current_exception());
+          }
```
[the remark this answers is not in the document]
lewissbaker: @jaredhoberock Do we have a process for collecting these sort of issues to be discussed?
Github issues on this repo?

wording.md
```diff
@@ -225,108 +247,175 @@ For some subexpressions `e` and `f`, let `E` be a type such that `decltype((e))`
...
+
+        void submit();
+
+    and that does not include a declaration of `execution::submit`. If the function selected by overload resolution does not submit the receiver object `r` via the sender `s`, the program is ill-formed with no diagnostic required.
+
+- Otherwise, `execution::start((new `_`submit-receiver`_`<S, R>{s,r})->state_)`, where _`submit-receiver`_
+is an implementation-defined class template equivalent to
+
+        template<class S, class R>
+        struct submit-receiver {
```
lewissbaker: "submit-receiver" is probably the wrong name for this since it isn't actually a receiver.

Perhaps "submit-state" or "submit-operation" would be better?
Then you could give `wrap` a more appropriate name (since it _is_ a receiver).

wording.md
```diff
@@ -227,108 +247,175 @@ For some subexpressions `e` and `f`, let `E` be a type such that `decltype((e))`
...
         private:
-          using invocable_type = std::remove_cvref_t<F>;
-          invocable_type f_;
-        public:
-          explicit as-receiver(invocable_type&& f) : f_(move_if_noexcept(f)) {}
-          explicit as-receiver(const invocable_type& f) : f_(f) {}
-          as-receiver(as-receiver&& other) = default;
-          void set_value() {
+          F f_;
+          void set_value() noexcept(is_nothrow_invocable_v<F&>) {
```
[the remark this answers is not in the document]
lewissbaker: The methods are rvalue-qualified in `submit-receiver::wrap`.
We should at least be consistent.

Keeping it rvalue-qualified would keep the door open for lvalue-qualified `set_value()` to be used for delivering non-terminal values in a stream/bulk situation.

wording.md
```diff
@@ -227,108 +247,175 @@ For some subexpressions `e` and `f`, let `E` be a type such that `decltype((e))`
...
+
+    and that does not include a declaration of `execution::start`.
+
+- Otherwise, `execution::start(O)` is ill-formed.
+
+#### `execution::submit`
+
+The name `execution::submit` denotes a customization point object.
+
+For some subexpressions `s` and `r`, let `S` be a type such that `decltype((s))` is `S` and let `R` be a type such that `decltype((r))` is `R`. The expression `execution::submit(s, r)` is ill-formed if `sender_to<S, R>` is not `true`. Otherwise, it is expression-equivalent to:
```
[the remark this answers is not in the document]
lewissbaker: The wording in that section is for a slightly different situation.
It starts with type, `T`, and then is defining an expression that has that type.
The form "Given a type X, let x be an expression such that decltype((x)) is X" makes sense for that situation.

Here we are starting with expression and giving a name to the type of that expression.
I think for this situation the form "Given an expression x, let X be decltype((x))" seems a more natural.

It's fairly minor in the scheme of things though.

wording.md
lewissbaker: I'm having trouble seeing where the recursion would be occurring here.
I'll have to get you to explain it to me :)

wording.md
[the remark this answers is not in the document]
jaredhoberock: @lewissbaker Yes, just file them as normal Github issues. I'll create a tag for in-committee wording review.

